PR #40, opened 2021-01-10
Replace Account and Path with WalletTools

passabilities:
### Account & Path => Keymanager API
With the update to the keymanager, the use of the Account and Path objects are no longer needed. 
Instead, I created wallet tools that sits on top of they keymanager that the code uses to derive public and private keys and addresses.

### Plugin Util Functions
In an attempt to keep the code modularized and functional, I created some plugin util functions that can be called anywhere in the codebase.


passabilities (commented):
rename folder to utxo engine

src/common/utxobased/plugin/utils.ts
```diff
@@ -0,0 +1,150 @@
...
+  args.keys[getXprivKey(args)]
+
+export const getXpub = (args: { keys: any, coin: string }): string =>
+  args.keys[getXpubKey(args)]
+
+export const getWalletCoinType = (args: { keys: any }): number =>
+  args.keys.coinType
+
+export const getWalletFormat = (args: { keys: any }): CurrencyFormat =>
+  args.keys.format
```
passabilities, line 82: return default if no exists

src/common/utxobased/plugin/utils.ts
```diff
@@ -0,0 +1,150 @@
...
+  `${coin}Xpub`
+
+export const getXpriv = (args: { keys: any, coin: string }): CurrencyFormatKeys =>
+  args.keys[getXprivKey(args)]
+
+export const getXpub = (args: { keys: any, coin: string }): string =>
+  args.keys[getXpubKey(args)]
+
+export const getWalletCoinType = (args: { keys: any }): number =>
+  args.keys.coinType
```
passabilities, line 79: return default if not exists

passabilities: `0`

src/common/utxobased/plugin/utils.ts
passabilities: `bip32`

src/common/utxobased/plugin/utils.ts
```diff
@@ -0,0 +1,150 @@
...
+    case BIP43PurposeTypeEnum.WrappedSegwit:
+      keys[getCurrencyFormatFromPurposeType(BIP43PurposeTypeEnum.Segwit)] = seedOrMnemonicToXPriv({
+        ...xprivArgs,
+        type: BIP43PurposeTypeEnum.Segwit
+      })
+      keys[getCurrencyFormatFromPurposeType(BIP43PurposeTypeEnum.WrappedSegwit)] = seedOrMnemonicToXPriv({
+        ...xprivArgs,
+        type: BIP43PurposeTypeEnum.WrappedSegwit
+      })
+      break
```
passabilities, line 129: get code from local changes

thehobbit85 (approved):

PR #41, opened 2021-01-10
seedOrMnemonicToXPriv to encode with base64

passabilities:
When a seed or mnemonic is passed in to the keymanager, it should be using the `base64` encoding when casting to a `Buffer`


thehobbit85 (commented):
Are you sure that we don't have a test for a "hex" seed that would break with this change?

thehobbit85 (approved):

PR #42, opened 2021-01-11
Noah/get fresh address

passabilities:
### Record and Return Fresh Address for Each Wallet Format Being Used
Because a wallet can be using 2 different paths (Segwit and WrappedSegwit) we have to watch and keep a record of the fresh indices for each. 

If a Segwit wallet is active, we will return the Segwit and WrappedSegwit (public) addresses. Otherwise, we just return the non Segwit (public) and legacy format, if available.


src/common/utxobased/plugin/makeUtxoEngine.ts
```diff
@@ -271,6 +274,10 @@ export async function makeUtxoEngine(config: EngineConfig): Promise<EdgeCurrency
         privateKeys
       })
 
+      for (const address of transaction.ourReceiveAddresses) {
```
thehobbit85, line 277: You shouldn't mark the address "used" at this stage. It should happen on "saveTx" so when that is implemented move it there.

src/common/utxobased/plugin/makeUtxoWalletTools.ts
```diff
@@ -70,23 +68,47 @@ export function makeUtxoWalletTools(config: WalletToolsConfig): UTXOPluginWallet
...
     addressToScriptPubkey(address: string): string {
-      return addressToScriptPubkey({
+      const args = {
         address,
-        addressType: getAddressTypeFromKeys(config),
         network,
         coin
-      })
+      }
+      try {
```
thehobbit85, line 87: I thought "addressToScriptPubkey" should already try and guess the "addressType" so why are you doing all this "try-catch"?

thehobbit85 (changes requested):

PR #49, opened 2021-02-01
Replace ProcessorTransaction

passabilities:
### Replace ProcessorTransaction with object helper functions
The class `ProcessorTransaction` has been removed and now only dealing with native object with helper functions.

### Start Listening to Emitted Messages on Engine Start
Added a function to start listening to emitted messages only once the engine is started and stop listening on engine stop.


thehobbit85:
I'm guessing this is irrelevant since we are restructuring the PR

PR #50, opened 2021-02-02
Block book notification tests

TheCharlatan:
To facilitate the tracking of an unconfirmed transaction an additional CountBase is created to keep track of the unconfirmed transactions. A count base was chosen to enable extension of its functionality into tracking non-matured transactions. In such a scenario each coin would have a defined maturity limit. This would also make it easier to detect and track transaction re-orgs.

If a new transaction is saved to the Transaction base with a block height of 0, its txid is added to count base at index 0.

Upon each incoming block notification all txids in the unconfirmed transaction count base are fetched. For each of them, their transaction data is re-fetched from Blockbook and compared with their existing saved transaction. If its block height is changed, the new transaction updates the old one and its txid is removed from the count base.

Blockbook does not expose an endpoint for block headers, so it's not possible to simply check for txid inclusion in them. Instead, the entire block can be fetched from Blockbook. It does not seem worth the extra bandwidth, especially compared to fetching just a couple of transactions. It might be interesting to eventually fetch the entire block when reaching a threshold of unconfirmed transactions.

### DIFF
--- a/src/common/utxobased/db/Models/baselet.ts
+++ b/src/common/utxobased/db/Models/baselet.ts
@@ -30,6 +30,13 @@ export const addressPathByMRUConfig: BaseletConfig<BaseType.CountBase> = {
   bucketSize: 100
 }
 
+export type TxIdsByConfirmations = string[]
+export const txIdsByConfirmationsConfig: BaseletConfig<BaseType.CountBase> = {
+  dbName: 'txIdByConfirmations',
+  type: BaseType.CountBase,
+  bucketSize: 1
+}
+
 export type ScriptPubKeysByBalance = {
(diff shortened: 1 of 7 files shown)


paullinator:
I have not looked at the code but based on your description above it looks like a countbase is the incorrect database type for tracking unconfirmed transactions. The problem with a countbase is that you can only put one transaction per entry. It's basically an array. The architecture designed, but not yet implemented, was to use a rangebase. Then you could put all unconfirmed transactions at the range value zero and easily find them. As well you could also put transactions into the range base on their actual blockheight and if you needed to drop a transaction that was actually confirmed, you could easily find them based on the block height.

PR #51, opened 2021-02-08
Update block height of unconfirmed transactions

TheCharlatan:
To facilitate the tracking of e.g. unconfirmed transactions an additional RangeBase is created to keep track of transaction id's by
block height.  A RangeBase was chosen to enable extension of its functionality into tracking non-matured transactions. In such a scenario each coin would have a maturity limit. This would also make it easier to detect and track transaction re-orgs.

If a new transaction is saved to the transaction base with a block height of 0, its txid is added to the range base with block height 0.

Upon each incoming block notification all txids in the block height range base with a block height of 0 are fetched. For each of them, their transaction data is re-fetched from Blockbook and compared with their existing saved transaction. If its block height is changed, the new transaction updates the old one and its txid is removed and re-added with the new block height in the range base.

Blockbook does not expose an endpoint for block headers, so it's not possible to simply check for txid inclusion in them. Instead, the entire block can be fetched from Blockbook, but it does not seem worth the extra bandwidth, especially compared to fetching just a couple of transactions. It might be interesting to eventually fetch the entire block when reaching a threshold of unconfirmed transactions.

Also adds a test for the notifications. A test for the full block height updating procedure with a dummy block book server will be done in another PR.

### DIFF
--- a/src/common/utxobased/db/Models/baselet.ts
+++ b/src/common/utxobased/db/Models/baselet.ts
@@ -29,6 +29,16 @@ export const addressPathByMRUConfig: BaseletConfig<BaseType.CountBase> = {
   bucketSize: 100
 }
 
+export interface TxIdsByBlockHeight {
+  [RANGE_ID_KEY]: string
+  [RANGE_KEY]: string
+}
+export const txIdsByBlockHeightConfig: BaseletConfig<BaseType.RangeBase> = {
+  dbName: 'txIdByConfirmations',
+  type: BaseType.RangeBase,
+  bucketSize: 100000
+}
+
 export type ScriptPubkeysByBalance = {
   [RANGE_ID_KEY]: string
   [RANGE_KEY]: string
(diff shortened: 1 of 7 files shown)


src/common/utxobased/db/Processor.ts
```diff
@@ -69,6 +71,14 @@ export interface Processor {
 
   updateAddressByScriptPubKey(scriptPubKey: string, data: Partial<IAddress>): void
 
+  insertAddressPathBySPubKey(scriptPubKey: string, addressPath: string): Promise<void>
```
thehobbit85, line 74: What is that? 
It looks like there is no implementation for this method.

thehobbit85 (changes requested):
Overall, looks great.
I really liked this way of just getting the transaction twice instead of the entire block.

I left one tiny comment that needs to be addressed.

src/common/utxobased/db/Processor.ts
TheCharlatan: Damn, this slipped in while testing some unrelated code. Removed.

thehobbit85 (approved):

thehobbit85:
Needs to resolve conflicts before it can be merged.

TheCharlatan:
Rebased.

thehobbit85 (approved):

PR #54, opened 2021-02-17
Update processor db return types

passabilities:
It slightly modifies the db return types and bucket sizes.


thehobbit85 (approved):

thehobbit85 (changes requested):
It is basically approved. 
But you need to fix the conflicts first.

thehobbit85 (approved):

PR #56, opened 2021-02-17
Resync blockchain

TheCharlatan:
This pull request builds largely on @passabilities original work, adding functionality to the current makeUtxoEngine `resyncBlockchain` function stub. All data in the baselets is now retained in an additional top-level disklet that is itself subdir'ed from the original passed in disklet. This should ensure that only the data of the resyncing blockchain is deleted.

To test its functionality, an address is added to the baselets and then removed with the new clearAll function.

### DIFF
--- a/src/common/utxobased/db/makeProcessor.ts
+++ b/src/common/utxobased/db/makeProcessor.ts
@@ -1,13 +1,20 @@
-import { BaseType, createCountBase, createHashBase, createRangeBase, openBase } from 'baselet'
+import {
+  BaseType,
+  clearMemletCache,
+  createCountBase,
+  createHashBase,
+  createRangeBase,
+  openBase
+} from 'baselet'
 import { CountBase } from 'baselet/src/CountBase'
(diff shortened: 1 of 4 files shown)


thehobbit85:
It's probably best to wait with this PR for now since there are some pretty big changes that needs to be merged that would effect the implementation of this in a major way.

src/common/utxobased/db/makeProcessor.ts
```diff
@@ -106,26 +157,28 @@ async function createOrOpen<T extends BaseType>(
...
+        return await createRangeBase(
+          disklet,
+          config.dbName,
+          config.bucketSize,
+          RANGE_KEY,
+          RANGE_ID_KEY
+        )
     }
   } catch (err) {
+    console.log('already exists', config)
```
thehobbit85, line 169: Remove this console.log.
As a general rule, don't use 'console' anywhere other than in the tests.
We will discuss how to properly log things in our meeting on Friday.

thehobbit85 (approved):

PR #70, opened 2021-02-26
Engine: Add json payment protocol options

TheCharlatan:
This is a port of the old plugin behavior to the new plugin. The old plugin had additional address sanity checks and converted addresses to legacy format. These additional steps are dropped. Some redundant arguments in create payment were also dropped.

### DIFF
--- a/src/common/utxobased/engine/makeUtxoEngine.ts
+++ b/src/common/utxobased/engine/makeUtxoEngine.ts
@@ -25,6 +25,7 @@ import { fetchMetadata, setMetadata } from '../../plugin/utils'
 import { fetchOrDeriveXprivFromKeys, getXprivKey } from './utils'
 import { IProcessorTransaction } from '../db/types'
 import { fromEdgeTransaction, toEdgeTransaction } from '../db/Models/ProcessorTransaction'
+import { createPayment, getPaymentDetails, sendPayment } from './paymentRequest'
 
 export async function makeUtxoEngine(config: EngineConfig): Promise<EdgeCurrencyEngine> {
   const {
@@ -35,7 +36,8 @@ export async function makeUtxoEngine(config: EngineConfig): Promise<EdgeCurrency
       walletLocalDisklet,
       walletLocalEncryptedDisklet,
       emitter
-    }
+    },
+    io
   } = config
 
   // Merge in the xpriv into the local copy of wallet keys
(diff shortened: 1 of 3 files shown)


src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -268,6 +295,18 @@ export async function makeUtxoEngine(config: EngineConfig): Promise<EdgeCurrency
         coin: currencyInfo.network,
         privateKeys
       })
+      const { edgeSpendInfo } = transaction.otherParams
+      const { paymentProtocolInfo } = edgeSpendInfo.otherParams
+      if (typeof paymentProtocolInfo !== 'undefined') {
+        const publicAddress = this.getFreshAddress({}).publicAddress
```
thehobbit85, line 301: This 'publicAddress' is never used and can be removed.

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -108,6 +110,27 @@ export async function makeUtxoEngine(config: EngineConfig): Promise<EdgeCurrency
     },
 
     async broadcastTx(transaction: EdgeTransaction): Promise<EdgeTransaction> {
+      const { otherParams } = transaction
+      const { paymentProtocolInfo } = otherParams
+
+      if (
+        typeof paymentProtocolInfo !== 'undefined' &&
```
thehobbit85, line 117: Avoid using 'typeof !== undefined'
Instead, use something like:
'otherParams?.paymentProtocolInfo?. payment != null && otherParams?.paymentProtocolInfo?. paymentUrl != null'

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -268,6 +295,18 @@ export async function makeUtxoEngine(config: EngineConfig): Promise<EdgeCurrency
         coin: currencyInfo.network,
         privateKeys
       })
+      const { edgeSpendInfo } = transaction.otherParams
+      const { paymentProtocolInfo } = edgeSpendInfo.otherParams
+      if (typeof paymentProtocolInfo !== 'undefined') {
```
thehobbit85, line 300: Avoid using 'typeof !== undefined'
Instead, use something like:
'paymentProtocolInfo != null'

thehobbit85 (changes requested):

src/common/utxobased/engine/makeUtxoEngine.ts
TheCharlatan: thank you :+1:

src/common/utxobased/engine/makeUtxoEngine.ts
TheCharlatan: Patched. Will use `!= null` from now.

thehobbit85 (approved):

PR #71, opened 2021-02-27
Fixing Typing Errors

passabilities:
It fixes errors caused by TypeScript.

### DIFF
--- a/src/common/utxobased/db/makeProcessor.ts
+++ b/src/common/utxobased/db/makeProcessor.ts
@@ -119,31 +119,42 @@ export async function makeProcessor(config: ProcessorConfig): Promise<Processor>
     emitter
   } = config
 
-  const [
-    scriptPubkeyByPath,
-    addressByScriptPubkey,
-    addressPathByMRU,
-    scriptPubkeysByBalance,
-    txById,
-    txIdsByBlockHeight,
-    txsByScriptPubkey,
-    txsByDate,
-    utxoById,
-    utxoIdsByScriptPubkey,
-    utxoIdsBySize
-  ] = await Promise.all([
+  const countBases = Promise.all([
     createOrOpen(disklet, scriptPubkeyByPathConfig),
-    createOrOpen(disklet, addressByScriptPubkeyConfig),
     createOrOpen(disklet, addressPathByMRUConfig),
+  ])
+  const rangeBases = Promise.all([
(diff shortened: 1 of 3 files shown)


src/common/utxobased/db/makeProcessor.ts
```diff
@@ -119,31 +119,42 @@ export async function makeProcessor(config: ProcessorConfig): Promise<Processor>
...
     createOrOpen(disklet, utxoIdsBySizeConfig)
   ])
+  const hashBases = Promise.all([
+    createOrOpen(disklet, addressByScriptPubkeyConfig),
+    createOrOpen(disklet, txByIdConfig),
+    createOrOpen(disklet, txsByScriptPubkeyConfig),
+    createOrOpen(disklet, utxoByIdConfig),
+    createOrOpen(disklet, utxoIdsByScriptPubkeyConfig)
+  ])
+  await Promise.all([countBases, rangeBases, hashBases])
```
thehobbit85, line 139: I understand you have issues with the types but doing the "await Promise.all" in this line and also doing individually "await" on "countBases/rangeBases/hashBases" in the next lines is redundent.

thehobbit85 (changes requested):
Fix the "Promise.all" issue

src/common/utxobased/db/makeProcessor.ts
passabilities: I don't have an issue with it, TypeScript throws errors when the array for the promise all is being destructed like this with more than 10 items:

image

thehobbit85: I understand that. You need to do something like this:

```
const [counts,  ranges , hashs ] = await Promise.all([countBases, rangeBases, hashBases])
const [
    scriptPubkeyByPath,
    addressPathByMRU
  ] = counts
  const [
    scriptPubkeysByBalance,
    txIdsByBlockHeight,
    txsByDate,
    utxoIdsBySize
  ] = await ranges
  const [
    addressByScriptPubkey,
    txById,
    txsByScriptPubkey,
    utxoById,
    utxoIdsByScriptPubkey
  ] = await hashs
```

passabilities: I mean that is the same thing I'm doing without creating additional variables

PR #74, opened 2021-02-27
Check Address Checked Progress

passabilities:
Adds the engine callback while processing address to keep track of the progress.

### DIFF
--- a/src/common/utxobased/engine/makeUtxoEngineState.ts
+++ b/src/common/utxobased/engine/makeUtxoEngineState.ts
@@ -23,7 +23,7 @@ import {
   validScriptPubkeyFromAddress
 } from './utils'
 import { makeMutexor, Mutexor } from './mutexor'
-import { BLOCKBOOK_TXS_PER_PAGE } from './constants'
+import { BLOCKBOOK_TXS_PER_PAGE, CACHE_THROTTLE } from './constants'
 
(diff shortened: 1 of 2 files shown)


src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -63,13 +63,22 @@ export function makeUtxoEngineState(config: UtxoEngineStateConfig): UtxoEngineSt
 
   return {
     async start(): Promise<void> {
+      let processedCount = 0
```
thehobbit85, line 66: You need to keep the 'ratio' value so you are able to throttle emitting the event.

Check this line in the old implementation for more detail:
https://github.com/EdgeApp/edge-currency-bitcoin/blob/698f4609a11ae8fd88ff3a8b53b1b3de40c1d0ec/src/engine/engineState.js#L511

thehobbit85 (changes requested):

src/common/utxobased/engine/makeUtxoEngineState.ts
passabilities: So the event is only emitted at max 4 times for each wallet?

thehobbit85 (approved):

thehobbit85 (approved):

PR #73, opened 2021-02-27
Refactor Engine State 

passabilities:
### Merge Code Paths
The engine state has been refactored which simplifies the flow of into smaller modular functions. By doing this, the code paths that are executed when processing and address for the first time and while an address is subscribed to for new transactions is the same.

PR #77, opened 2021-03-03
Sweep private key

TheCharlatan:
Based on #73 and #74.

Sweep private key is implemented in the UtxoEngine's sweepPrivateKey function. To facilitate fetching the to be spent utxos, a temporary engine state is created. The UtxoEngineState emitter is in turn used as a callback to notify when all addresses have been scanned. The callback then resolves the Promise of the sweepPrivateKey function. On success, it returns an unsigned EdgeTransaction.

To support sweepPrivateKey, additional changes were done in the wallet tools to support wif keys and the keymanager to support creating transactions with pre-selected UTXOs.

### DIFF
--- a/src/common/utxobased/engine/makeUtxoEngine.ts
+++ b/src/common/utxobased/engine/makeUtxoEngine.ts
@@ -252,7 +252,10 @@ export async function makeUtxoEngine(
       return !!addressData?.used
     },
 
-    async makeSpend(edgeSpendInfo: EdgeSpendInfo, options?: TxOptions): Promise<EdgeTransaction> {
+    async makeSpend(
+      edgeSpendInfo: EdgeSpendInfo,
+      options?: TxOptions
+    ): Promise<EdgeTransaction> {
       const targets: MakeTxTarget[] = []
       const ourReceiveAddresses: string[] = []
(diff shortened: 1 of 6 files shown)


src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -344,9 +350,83 @@ export async function makeUtxoEngine(config: EngineConfig): Promise<EdgeCurrency
...
+
+          this.makeSpend(spendInfo, options)
+            .then(tx => success(tx))
+            .catch(e => failure(e))
+        }
+      })
+
+      config.options.emitter = tmpEmitter
+      // hack to not overflow the wallet tools private key array
+      config.currencyInfo.gapLimit = privateKeys.length + 1
```
thehobbit85, line 420: This might cause problems later on.
I saw you set it back when on 'ADDRESSES_CHECKED' when the ratio is 1 but we can't really know how long it takes, or that it would actually get there.

thehobbit85 (changes requested):
Rethink how to implement the private key array 'overflow' problem

thehobbit85 (approved):

thehobbit85 (approved):

PR #78, opened 2021-03-05
DumpData

passabilities:
### Dump Data
Added more info about the wallet and stored data such as all addresses, transactions and utxos.

Example output:
image


thehobbit85 (changes requested):
The goal of "dumpData" is to help debugging runtime issues, and not to get a nicely parsed version of the wallet's data.

Try and return the data as close as possible to what we actually save on disk.

thehobbit85 (approved):

PR #79, opened 2021-03-10
Engine Fees

passabilities:
### Fee Structure
Each currency info object has default fee values to use. Edge has their own server that returns some standard fee values for currencies that overwrite the defaults. Vendors, such as Earn.com and Mempool.space, are used to get more accurate up-to-date fee values on an interval that overwrites Edge's and the default values. 

These values are fetched and then cached to disk. On the next load, the engine fetches the cached values to then overwrite the currency defaults before it attempts to fetch fee values from Edge or any vendors.

### Calculate Rate
The previously implemented function, `calculateFeeRate`, has been replaced with the most recent implementation from [edge-bitcoin-plugin](https://github.com/EdgeApp/edge-currency-bitcoin/blob/d522ed1428/src/engine/miningFees.js#L195).


src/common/fees.ts
```diff
@@ -49,39 +49,40 @@ export const makeFees = async (config: MakeFeesConfig): Promise<Fees> => {
...
 
     await cacheFees(memlet, fees)
   }
 
-  let vendorIntervalId: NodeJS.Timeout
+  let edgeIntervalId: NodeJS.Timeout
 
   return {
     async start(): Promise<void> {
+      const vendorFees = await fetchFeesFromVendor({
```
passabilities, line 64: fetch fees from edge,
fetch fees from vendors - if ANY of them return then continue

passabilities (commented):
- separate vendor fetching from calculating.
- put calculating functionality in own file
- import test files from old bitcoin plugin

passabilities:
build on top of changes made during PR review

thehobbit85 (changes requested):
Address Comments and rebase/resolve conflicts

thehobbit85 (approved):

PR #83, opened 2021-03-14
Signed Transaction Id

passabilities:
### Signed Transaction Id
After signing a transaction, we return the transaction which is in turn passed to `engine.saveTransaction()`. Previously, the `txid` was not being set. Since in order to calculate the txid, the full transaction needs to be signed, we should set the id when we set the `signedTx` value on the `EdgeTransaction` as well.

Also going from an `EdgeTransaction` to `ProcessorTransaction` was not preserving the `signedTx` or `hex` value.

### DIFF
--- a/src/common/utxobased/engine/makeUtxoEngine.ts
+++ b/src/common/utxobased/engine/makeUtxoEngine.ts
@@ -27,6 +27,7 @@ import { IProcessorTransaction } from '../db/types'
 import { fromEdgeTransaction, toEdgeTransaction } from '../db/Models/ProcessorTransaction'
 import { createPayment, getPaymentDetails, sendPayment } from './paymentRequest'
 import { makeMutexor } from './mutexor'
+import { UTXOTxOtherParams } from './types'
 
 export async function makeUtxoEngine(config: EngineConfig): Promise<EdgeCurrencyEngine> {
   const {
@@ -248,7 +249,7 @@ export async function makeUtxoEngine(config: EngineConfig): Promise<EdgeCurrency
       }
 
       let nativeAmount = '0'
-      for (const output of tx.psbt.txOutputs) {
+      for (const output of tx.outputs) {
         const scriptPubkey = output.script.toString('hex')
         const own = await processor.hasSPubKey(scriptPubkey)
         if (!own) {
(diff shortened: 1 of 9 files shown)


src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -1081,5 +1083,5 @@ export async function signTx(args: SignTxArgs): Promise<string> {
   console.log('fee:', psbt.getFee())
   console.log('feeRate:', psbt.getFeeRate())
   console.log('-----------------------------------------')
-  return psbt.extractTransaction().toHex()
+  return tx
```
thehobbit85, line 1086: Instead of returning the TX, return the {id, hex}

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -261,11 +261,11 @@ export async function makeUtxoEngine(config: EngineConfig): Promise<EdgeCurrency
       return {
         ourReceiveAddresses,
         otherParams: {
-          psbt: tx.psbt.toBase64(),
+          psbtBase64: tx.psbt.toBase64(),
```
thehobbit85, line 264: instead of using other params and base64, use signedTx and hex

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -261,11 +261,11 @@ export async function makeUtxoEngine(config: EngineConfig): Promise<EdgeCurrency
...
           edgeSpendInfo
         },
         currencyCode: currencyInfo.currencyCode,
-        txid: '',
+        txid: tx.psbt.toHex(),
```
thehobbit85, line 268: keep at ''

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -286,24 +286,29 @@ export async function makeUtxoEngine(config: EngineConfig): Promise<EdgeCurrency
     },
 
     async signTx(transaction: EdgeTransaction): Promise<EdgeTransaction> {
-      const { psbt } = transaction!.otherParams!
-      const inputs = bitcoin.Psbt.fromBase64(psbt).txInputs
-      const privateKeys = await Promise.all(inputs.map(async ({ hash, index }) => {
-        const txid = Buffer.isBuffer(hash) ? hash.reverse().toString('hex') : hash
+      const { psbtBase64 } = transaction!.otherParams!
```
thehobbit85, line 289: use the "signedTx" param

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -286,24 +286,29 @@ export async function makeUtxoEngine(config: EngineConfig): Promise<EdgeCurrency
...
+      const psbt = bitcoin.Psbt.fromBase64(psbtBase64)
```
thehobbit85, line 290: use from hex

thehobbit85 (changes requested):
Changes requested

thehobbit85 (approved):

PR #92, opened 2021-03-22
Display Public Seed

passabilities:
Implements the UTXO Engine function `getDisplayPublicSeed` function.


thehobbit85 (approved):

thehobbit85 (changes requested):
You need to handle the case where you have more than 1 xpub.

thehobbit85 (approved):

PR #93, opened 2021-03-22
Display Private Key

passabilities:
It implements the UTXO Engine function `getDisplayPrivateKey` function.


thehobbit85 (changes requested):
As discussed, you should return the seed and not the xpriv

thehobbit85 (approved):

PR #95, opened 2021-03-29
EngineEmitter

passabilities:
It creates an `Emitter` object that is specific to the use cases of the engine events.

### DIFF
--- a/src/common/utxobased/network/BlockBook.ts
+++ b/src/common/utxobased/network/BlockBook.ts
@@ -1,6 +1,6 @@
 import { EdgeConsole, EdgeTransaction } from 'edge-core-js'
 
-import { Emitter, EmitterEvent } from '../../plugin/types'
+import { EngineEmitter, EngineEvent } from '../../plugin/makeEngineEmitter'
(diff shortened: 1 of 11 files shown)


src/common/utxobased/network/BlockBook.ts
```diff
@@ -9,17 +9,20 @@ export interface INewTransactionResponse {
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [id: string]: (value: any) => void
 }
 
 interface IWsMessage {
   id: string
   method: string
-  params?: object
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  params?: any
```
passabilities, line 20: follow up

src/common/utxobased/network/BlockBook.ts
```diff
@@ -9,17 +9,20 @@ export interface INewTransactionResponse {
...
 }
 
 interface IWsResponse {
   id: string
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   data: any
```
passabilities, line 26: follow up

PR #100, opened 2021-03-29
Address Data Not Found

passabilities:
If an address data was not found while searching for the last used, save the new address.


samholmes:
Closing because we've lost tracking and context on this work. It appears to be irrelevant now; the function it modifies no longer exists.

PR #108, opened 2021-04-09
Logging: Use plugin EdgeLog object

TheCharlatan:


src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -66,7 +66,7 @@ export function makeUtxoEngineState(
       processor
     })
     const percent = processedCount / totalCount
-    config.io.console.info('processed', percent)
+    log('processed', percent)
```
thehobbit85, line 69: remove this log

thehobbit85 (changes requested):
small change

PR #131, opened 2021-05-14
Fix resync behavior

TheCharlatan:
Fixes to the metadata cache to update the balance correctly when starting a resync and the serverCache to clear any lingering data and connections.

### DIFF
--- a/src/common/plugin/makeMetadata.ts
+++ b/src/common/plugin/makeMetadata.ts
@@ -13,7 +13,7 @@ interface MetadataConfig {
 }
 
 export interface Metadata extends LocalWalletMetadata {
-  clear: () => Promise<void>
+  clear: (currencyCode: string) => Promise<void>
 }
 
 export const makeMetadata = async (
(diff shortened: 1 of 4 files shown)


src/common/plugin/makeMetadata.ts
```diff
@@ -51,9 +51,14 @@ export const makeMetadata = async (
     get lastSeenBlockHeight() {
       return cache.lastSeenBlockHeight
     },
-    clear: async () => {
+    clear: async (currencyCode: string) => {
       await memlet.delete(metadataPath)
-      cache = await fetchMetadata(memlet)
+      cache = await resetMetadata(memlet)
+      emitter.emit(
+        EngineEvent.WALLET_BALANCE_CHANGED,
```
TheCharlatan, line 58: Instead of calling it on clear, emit a 0 balance at start engine.

TheCharlatan: Moved to UtxoEngine.

thehobbit85 (approved):

PR #136, opened 2021-05-22
Change Maps and Sets to Arrays and Objects

TheCharlatan:


src/common/utxobased/engine/makeServerStates.ts
```diff
@@ -68,8 +75,10 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
     log(`${uri} ** Connected **`)
   })
   emitter.on(EngineEvent.CONNECTION_CLOSE, (uri: string, error?: Error) => {
-    connections.delete(uri)
-    serverStates.delete(uri)
+    // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+    delete connections[uri]
```
thehobbit85, line 79: It might be safer to do it like this:

```
const removeItem = (obj, key) => {
   const { [key]: _, ...rest } = obj
   return rest
}

connections = removeItem(connections, uri)
serverStates = removeItem(serverStates, uri)
```

thehobbit85 (changes requested):
The comment I wrote is relevant to every place you use `delete`

src/common/utxobased/engine/makeServerStates.ts
TheCharlatan: I can create a generic like so:
```
interface GenericObject<T> {
  [key: string]: T
}

export const removeItem = <T>(
  obj: GenericObject<T>,
  key: string
): GenericObject<T> => {
  const { [key]: _, ...rest } = obj
  return rest
}
```
But isn't this super slow? We're essentially copying and then replacing the entire object.

thehobbit85 (approved):

PR #151, opened 2021-06-10
Implement cleaners

TheCharlatan:


src/common/plugin/makeMetadata.ts
```diff
@@ -78,7 +78,7 @@ const fetchMetadata = async (memlet: Memlet): Promise<LocalWalletMetadata> => {
     const dataStr = await memlet.getJson(metadataPath)
     return JSON.parse(dataStr)
   } catch {
-    return await resetMetadata(memlet)
+    return asLocalWalletMetadataCleaner(await resetMetadata(memlet))
```
thehobbit85, line 81: You don't need to clean "internal" functions that returns a known type.
Cleaners are used when interacting with external data like other servers or databases.

src/common/fees/makeFees.ts
```diff
@@ -155,7 +162,7 @@ const fetchFeesFromVendor = async (
       processor: processEarnComFees
     })
     if (earnComFees != null) {
-      return earnComFees
+      return asFeeRatesCleaner(earnComFees)
```
thehobbit85, line 165: Notice that cleaners "throw" in case they fail. 
So you should probably either pass the cleaner to the "fetchFees" function and call it inside.

src/common/utxobased/db/makeProcessor.ts
```diff
@@ -182,20 +190,22 @@ export async function makeProcessor(
     },
 
     async getUsedAddress(scriptPubkey: string): Promise<boolean> {
+      const cleaner = asBoolean
       const [used] = await baselets.all.usedFlagByScriptPubkey.query('', [
         scriptPubkey
       ])
-      return used
+      return cleaner(used)
```
thehobbit85, line 197: Instead of what you did here, you should have something like this:
```
const doc = await baselets.all.usedFlagByScriptPubkey.query('', [
   scriptPubkey
])
return asUsed(doc) // This cleaner would verify the doc and return just the used if it's valid
```

This comment is relevant to almost all your usages of cleaners in this file

src/common/utxobased/db/makeProcessor.ts
```diff
@@ -182,20 +190,22 @@ export async function makeProcessor(
...
     },
 
     async fetchScriptPubkeyByPath(
       path: AddressPath
     ): Promise<ScriptPubkeyByPath> {
+      const cleaner = asOptional(asString)
```
thehobbit85, line 203: See comment above

thehobbit85 (changes requested):
See comments

src/common/plugin/types.ts
```diff
@@ -61,11 +61,27 @@ export interface FeeRates {
...
 
+export const asSimpleFeeSettingsCleaner = asObject({
+  standardFeeLowAmount: asString,
+  standardFeeHighAmount: asString,
+  lowFee: asString,
+  standardFeeLow: asString,
+  standardFeeHigh: asString,
+  highFee: asString
+})
+
```
samholmes, line 84: nit: Avoid using "Cleaner" in the name of a cleaner. No need to have the word there when the "as" naming convention is enough to signal it's a cleaner. ;)

samholmes (commented):
Some typos to fix in the 'Cleaners: Implement network cleaner' commit message:

```
Add a cleaner for subscription notification messages (new block and new
transaction) and the ServerInfo, Utxos, Transactions and
TransactionBroadcast response messages.

The implementation of the transaction cleaner brought to light that the
currently definition of the transaction response type is buggy. It
contained fields not present in the response, and not the complete set
either.

The response for an address message is a mix type, that could not be
easily implemented by the cleaner. Further investigation on its
implementation is required.
```

TheCharlatan:
Fixed the commit message typos.

src/common/utxobased/network/BlockBook.ts
```diff
@@ -1,9 +1,16 @@
...
 import { EngineEmitter, EngineEvent } from '../../plugin/makeEngineEmitter'
 import {
   addressMessage,
   addressUtxosMessage,
+  asAddressUtxosCleaner,
+  asINewBlockResponseCleaner,
+  asINewTransactionResponseCleaner,
+  asIServerInfoCleaner,
+  asITransactionBroadcastResponseCleaner,
+  asITransactionCleaner,
```
samholmes, line 13: Remove the word `Cleaner` again here. This way the cleaner for `ITransactionCleaner` is simply `asITransactionCleaner`

src/common/utxobased/db/makeProcessor.ts
```diff
@@ -793,7 +811,12 @@ const fetchAddressesByScriptPubkeys = async (
 
   // Short circuit query to the database
   if (scriptPubkeys.length === 0) return []
-  return await tables.addressByScriptPubkey.query('', scriptPubkeys)
+  const cleaner = asArray(asOptional(asIAddressCleaner))
+  return cleaner(
+    await tables.addressByScriptPubkey.query('', scriptPubkeys)
+    // allow any for cleaned string literal type
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ) as any
```
samholmes, line 823: I'm not seeing the reason for `as any` type assertion here. Please explain.

src/common/utxobased/db/makeProcessor.ts
```diff
@@ -1225,13 +1252,14 @@ const fetchTxIdsByBlockHeight = async (
...
   const result = await tables.txIdsByBlockHeight.query('', fromBlock, toBlock)
+  const cleaner = asString
 
   return (
     result
       // RangeBase returns values with lowest value first
       .reverse()
       // Return array of just the IDs
-      .map(({ [RANGE_ID_KEY]: id }) => id)
+      .map(({ [RANGE_ID_KEY]: id }) => cleaner(id))
```
samholmes, line 1266: Why not just `asString(id`?

src/common/fees/makeFees.ts
```diff
@@ -110,11 +111,17 @@ export const makeFees = async (config: MakeFeesConfig): Promise<Fees> => {
...
+  const data = await memlet
     .getJson(FEES_PATH)
     // Return the simple fees settings from currency info by default
     .catch(() => currencyInfo.simpleFeeSettings)
+  try {
+    return asSimpleFeeSettingsCleaner(data)
+  } catch (_) {
+    return
+  }
+}
```
samholmes, line 124: This is an opportunity to use `asMaybe`

test/common/fees/fees.spec.ts
```diff
@@ -39,6 +40,9 @@ describe('fees', function () {
         log: fakeLog
       })
 
+      // make typescript happy with this typecheck
+      if (fees.fees == null) fail()
```
samholmes, line 44: Include message parameter for `fail()`. It'll be handy later

samholmes (changes requested):
1. Nit on naming the cleaners
2. I'm concerned with the over-use of disabling no-explicit-any eslint rule; I feel like more justification is needed for this, or at least an explanation to easy my paranoia.

src/common/utxobased/db/makeProcessor.ts
TheCharlatan: There is a comment in the line above explaining it - is it not clear enough? I have not been able to get cleaners to work with either enums or like in this case string literals in the time I spent with them so far.

src/common/utxobased/db/makeProcessor.ts
TheCharlatan: I don't like the nesting syntax, e.g. `cleaner(cleaner(cleaner))(something)`, so I put the cleaner declaration on a separate line. Everywhere, no reason to do it, so I changed it.

src/common/fees/makeFees.ts
TheCharlatan: Fixed in: https://github.com/EdgeApp/edge-currency-plugins/pull/151/commits/38e026c454894fc1740a4ffa7838fa37f027209a

src/common/utxobased/db/makeProcessor.ts
TheCharlatan: Fixed in: https://github.com/EdgeApp/edge-currency-plugins/pull/151/commits/64988911484b0708bb85979760f61b270b35e194

test/common/fees/fees.spec.ts
TheCharlatan: Fixed in https://github.com/EdgeApp/edge-currency-plugins/pull/151/commits/6a3d20c541036a50c13f1ad58532e19d8292f5cf

src/common/utxobased/db/makeProcessor.ts
samholmes: Here's an example on how to resolve the issue: https://github.com/EdgeApp/edge-currency-plugins/commit/97f0c0ae4b863af34649d73b3622d466b3f6bbb0 

I'll add something that I originally overlooked: all cleaners that have a separate static type definition should include that static type as the type variable/argument to the `asObject` cleaner in order to tightly couple the two definitions and receive errors from the type-checker if the two definitions misalign.

src/common/utxobased/network/BlockBook.ts
samholmes: Tip: if you're using vscode you can secondary/right click on the identifier and choose Rename Symbol to easily refactor your code when it comes to name changes. I made this into a keyboard short-cut.


Screen Shot 2021-06-25 at 12 03 40

src/common/plugin/types.ts
```diff
@@ -83,3 +84,8 @@ export interface LocalWalletMetadata {
   balance: string
   lastSeenBlockHeight: number
 }
+
+export const asLocalWalletMetadataCleaner = asObject({
+  balance: asString,
+  lastSeenBlockHeight: asNumber
+})
```
samholmes, line 107: As mentioned, we want to either include `LocalWalletMetadata` static type in the cleaner's definition, or we want to derive the static type from the cleaner in order to maintain code coupling:

Option 1:
```ts
export const asLocalWalletMetadata = asObject<LocalWalletMetadata>({
  balance: asString,
  lastSeenBlockHeight: asNumber
})
```

Option 2 (Preferred/DRY):
```ts
export type LocalWalletMetadata = ReturnType<typeof asLocalWalletMetadata>
export const asLocalWalletMetadata = asObject({
  balance: asString,
  lastSeenBlockHeight: asNumber
})
```

This rule should be applied to all cleaner definitions in this PR.

samholmes (changes requested):

src/common/plugin/types.ts
TheCharlatan: This is nice, will apply everywhere!

src/common/utxobased/network/BlockBook.ts
TheCharlatan: I saved it for last so don't pollute the changes with renames.

src/common/plugin/types.ts
TheCharlatan: Done in the newest batch of fixup commits.

src/common/utxobased/db/makeProcessor.ts
TheCharlatan: I removed all the any type assertions in https://github.com/EdgeApp/edge-currency-plugins/pull/151/commits/c350510589e93b090a3f5bd819fa4015e685bc23 . Also implemented a custom cleaner for the enum type.

samholmes:
Replaced by https://github.com/EdgeApp/edge-currency-plugins/pull/183

PR #152, opened 2021-06-11
Add block book servers

TheCharlatan:
Fixup to the previous pull request

### DIFF
--- a/src/common/plugin/pluginState.ts
+++ b/src/common/plugin/pluginState.ts
@@ -27,7 +27,7 @@ interface JsonObject {
 
 export interface CurrencySettings {
   customFeeSettings: string[]
-  electrumServers: string[]
+  blockBookServers: string[]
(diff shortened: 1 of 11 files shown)


src/common/plugin/pluginState.ts
```diff
@@ -99,7 +99,7 @@ export class PluginState extends ServerCache {
     // TODO: Change the below line to blockbook once available
```
samholmes, line 99: Should this line be removed now?

samholmes (approved):
Small nit if applicable.

src/common/plugin/pluginState.ts
TheCharlatan: i removed it.

samholmes: Looks good. Apply the fixup commit, and then merge.

PR #167, opened 2021-08-13
New processor skeleton

TheCharlatan:


src/common/utxobased/db/NewModels/ProcessorTransaction.ts
```diff
@@ -0,0 +1,63 @@
+import { EdgeTransaction } from 'edge-core-js'
```
samholmes, line 1: This file is a dup of `src/common/utxobased/db/Models/ProcessorTransaction.ts`; we can just use the original file. 

Perhaps:

1. Mark each old file as deprecated with an "Old" suffix (i.e. `db/makeProcessorOld.ts` `db/Models/baseletsOld.ts`)
2. Rename "Models" folder to "models" (let's keep the case consistent).
3. Create the new files without the old suffixes
4. Later switch to the new files by remove the "Old" suffix from each import statement along with any implementation changes; then we can delete the old files within a single commit.

The alternative is to use a "New" suffix for new files and then _add_ the "New" suffix when updating the dependent modules, but this will require an extra rename commit from with "New" to without at a later point. However, I think this does worse of a job to communicate which files are deprecated while we progress.

src/common/utxobased/db/NewModels/baselet.ts
```diff
@@ -0,0 +1,86 @@
+import { BaseType } from 'baselet'
```
samholmes, line 1: A lot of this code is dup. Seems like there's less merit to renaming the old file (`baseletsOld.ts`) and rather we could just label certain declarations here with a `// deprecated` and `// deprecated (use LastUsedByFormatPath)` comments. this way we can clearly see which parts of the API is new and which parts will be removed/replaced.

src/common/utxobased/db/makeNewBaselets.ts
```diff
@@ -0,0 +1,189 @@
+import {
```
samholmes, line 1: I feel like we should do the same "bridging" of the API change here; that is add/rename any changes and annotate deprecated old code (soon to be dead and removed).

samholmes (changes requested):
In summary, my feedback is to make it clear how we bridge the changes from the existing processor API state to the new state with changes to commit structure and how we add new code and deprecate old code (soon to be removed code).

src/common/utxobased/db/makeNewBaselets.ts
```diff
@@ -0,0 +1,189 @@
...
+      return await fn(blockHeightBases)
+    },
+
+    all: {
+      ...addressBases,
+      ...txBases,
+      ...utxoBases,
+      ...blockHeightBases
+    }
+  }
```
samholmes, line 188: This was badly broken in my editor, and then I realized we really shouldn't be duplicated the type information within the implementation (`Baselets` should contain all the type information we need to type this return value). However, TypeScript wont be happy without any type annotations per method, the minimum I was able to achieve is:

```ts
return {
    async address(fn): Promise<ReturnType<typeof fn>> {
      return await fn(addressBases)
    },

    async tx(fn): Promise<ReturnType<typeof fn>> {
      return await fn(txBases)
    },

    async utxo(fn): Promise<ReturnType<typeof fn>> {
      return await fn(utxoBases)
    },

    async block(fn): Promise<ReturnType<typeof fn>> {
      return await fn(blockHeightBases)
    },

    all: {
      ...addressBases,
      ...txBases,
      ...utxoBases,
      ...blockHeightBases
    }
  }
```

samholmes (changes requested):
More improvements here.

src/common/utxobased/db/NewModels/ProcessorTransaction.ts
TheCharlatan: I disagree with this approach. We'd have to change any importing files to first point to the suffixed old, and then to the new file, instead of directly pointing to the new file. This feels much uncleaner to me than adding new files, then rename them and point the rest of the code at their new interfaces.

src/common/utxobased/db/newProcessor.ts
```diff
@@ -5,7 +5,7 @@ import { EdgeGetTransactionsOptions } from 'edge-core-js'
 import { AddressPath } from '../../plugin/types'
 import { IAddress, IProcessorTransaction, IUTXO } from './types'
 
-/* Transation table interfaces */
+/* Transaction table interfaces */
```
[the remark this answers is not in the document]
TheCharlatan: Done.

src/common/utxobased/db/newProcessor.ts
```diff
@@ -118,5 +127,106 @@ export interface Processor {
...
+    },
+
+    async fetchBlockHash(height: number): Promise<string[]> {
+      return await baselets.block(
+        async tables =>
+          await tables.blockHashByBlockHeight.query('', [height.toString()])
+      )
+    }
+  }
+  return processor
```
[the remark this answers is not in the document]
TheCharlatan: Implementation now sits in https://github.com/EdgeApp/edge-currency-plugins/pull/167/commits/9363ccdb3aaeffab0b9fd9f6565ca912c28b5e5d

src/common/utxobased/db/NewModels/baselet.ts
TheCharlatan: Yes, this looks cleaner to me as well. I chose this approach in the second commit of the pull request now: https://github.com/EdgeApp/edge-currency-plugins/pull/167/commits/8921fdd949d4f7718b89a151e7b63248911b1e2c

src/common/utxobased/db/makeNewBaselets.ts
TheCharlatan: Used what you suggested here.

samholmes (approved):
This looks good.

PR #168, opened 2021-08-16
New processor address processing

TheCharlatan:


src/common/utxobased/db/newProcessor.ts
```diff
@@ -183,31 +184,178 @@ export async function makeNewProcessor(
...
+      }
+      // fetch addresses by provided scriptPubkey
+      if (scriptPubkey != null) {
+        await baselets.address(async tables => {
+          const [address] = await tables.addressByScriptPubkey.query('', [
+            scriptPubkey
+          ])
+          addresses.push(address)
+        })
+      }
```
samholmes, line 353: This will only return `[IAddress] | [IAddress, IAddress]`? I wonder why we don't go with `fetchAddress(args: FetchAddressArgs): Promise<IAddress>` and then call twice at the call site; this way the caller knows which address returned is for which query argument (path or scriptPubkey).

test/common/utxobased/db/NewProcessor.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+    assertNumAddresses(2)
+    const [processorAddress4] = await processor.fetchAddresses({
+      scriptPubkey: 'testscriptpubkey3'
+    })
+    expect(processorAddress4.scriptPubkey).to.eqls(address4.scriptPubkey)
+    const lastUsed = await processor.lastUsedIndexByFormatPath({
+      format: 'bip44',
+      changeIndex: 0
+    })
+    expect(lastUsed).to.eql(1)
```
samholmes, line 128: Why not refactor your `assertLastUsedByFormatPathEmpty(): void` into a `assertLastUsedByFormatPath(toBe: number | undefined): void` and use it for this check as well?

test/common/utxobased/db/NewProcessor.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+    try {
+      await processor.updateAddress({
+        scriptPubkey: 'testscriptpubkey2',
+        data: { lastQuery: 0 }
+      })
+    } catch (err) {
+      threw = true
+      expect(err.message).to.eqls('Cannot update address that does not exist')
+    }
+    expect(threw).to.be.true
```
samholmes, line 164: Echo last comment

test/common/utxobased/db/NewProcessor.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+    assertNumAddresses(1)
+    const [processorAddress3] = await processor.fetchAddresses({
+      scriptPubkey: 'testscriptpubkey1'
+    })
+    expect(processorAddress3.scriptPubkey).to.eqls('testscriptpubkey1')
+    const lastUsed = await processor.lastUsedIndexByFormatPath({
+      format: 'bip44',
+      changeIndex: 0
+    })
+    expect(lastUsed).to.eql(0)
```
samholmes, line 192: Echo last comment

samholmes (changes requested):

src/common/utxobased/db/newProcessor.ts
```diff
@@ -35,7 +36,7 @@ interface UpdateAddressArgs {
 }
 
 interface FetchAddressArgs {
-  path?: Omit<AddressPath, 'addressIndex'> | AddressPath
+  path?: AddressPath
```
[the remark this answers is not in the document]
TheCharlatan: Oversight by me, I only though of fetching the last used address by the path prefix, but not by the full path. We need to be able to do both.

test/common/utxobased/db/NewProcessor.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+      changeIndex: 0
+    })
+    expect(result).to.be.undefined
+  }
+
+  it('empty address baselets', async () => {
+    const storage = {}
+    const disklet = makeMemoryDisklet(storage)
+
+    processor = await makeNewProcessor({ disklet })
```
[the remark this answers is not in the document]
TheCharlatan: Done.

test/common/utxobased/db/NewProcessor.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+    let threw = false
+    try {
+      await processor.insertAddress(address3)
+    } catch (err) {
+      expect(err.message).to.equal(
+        'Attempted to save address with an existing path, but different script pubkey'
+      )
+      threw = true
+    }
+    expect(threw).to.be.true
```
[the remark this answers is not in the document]
TheCharlatan: I did not find a canonical way to deal with async functions that throw. As far as I can tell  `assert.throws` cannot deal with async.

src/common/utxobased/db/newProcessor.ts
TheCharlatan: Yeah, seems cleaner to me as well.

test/common/utxobased/db/NewProcessor.spec.ts
TheCharlatan: Done.

src/common/utxobased/db/newProcessor.ts
```diff
@@ -197,6 +202,78 @@ export async function makeNewProcessor(
...
+              addressPathToPrefix(address.path),
+              address.path.addressIndex,
+              address.scriptPubkey
+            )
+          }
+
+          // Only update the used flag if one was given and is true
+          if (address.used != null && address.used && !address.used) {
+            address.used = true
+          }
```
samholmes:
```ts
if (address.used === true && existingAddress.used !== true) {
  existingAddress.used = true
}
```

samholmes (approved):
Add the fixup to the tests in 29989ed528f7e31a2b5112977d3567dc9d6b95b7

PR #169, opened 2021-08-20
New processor transaction

TheCharlatan:


src/common/utxobased/db/newProcessor.ts
```diff
@@ -161,26 +194,160 @@ export async function makeNewProcessor(
...
+              )
+            }
+            const txIdsByOptions = await txData
+              .reverse()
+              .map(({ [RANGE_ID_KEY]: id }) => id)
+
+            const txsByOptions = await tables.txById.query('', txIdsByOptions)
+            // Make sure only existing transactions are included
+            txs.push(...txsByOptions.filter(tx => tx != null))
+          } catch {}
```
samholmes, line 347: Fix silent fail here too

test/common/utxobased/db/NewProcessor.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+import {
+  IProcessorTransaction,
+  ITransactionInput,
+  ITransactionOutput
+} from '../../../../src/common/utxobased/db/types'
+
+chai.should()
+
+describe('Processor transactions tests', () => {
+  let processor: NewProcessor
```
samholmes, line 19: Don't share `processor` state between test if not necessary

test/common/utxobased/db/NewProcessor.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+      inputs: [input1],
+      outputs: [output1, output2],
+      ourIns: [],
+      ourOuts: [],
+      ourAmount: '0'
+    }
+
+    await processor.insertTransaction({
+      tx: transaction1,
+      scriptPubkey: 'pubkeyout1'
```
samholmes, line 71: Should use `output1.scriptPubkey` instead of duplicating the string? The same goes for other strings below; the reason for this is to make sure that if a string ever changes in the spec, it wont break and require many changes; it also voids the possibility of typos.

test/common/utxobased/db/NewProcessor.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+    const [txByBlockHeight1] = await processor.fetchTransactions({
+      blockHeight: 1
+    })
+    expect(txByBlockHeight1.blockHeight).to.eqls(1)
+
+    // insert the same transaction, but with a script pubkey referencing an input
+    await processor.insertTransaction({
+      tx: transaction1,
+      scriptPubkey: 'pubkeyin1'
+    })
```
samholmes, line 87: We can insert to update?

test/common/utxobased/db/NewProcessor.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    await processor.insertTransaction({
+      tx: transaction1,
+      scriptPubkey: 'pubkeyout1'
+    })
+    await processor.insertTransaction({
+      tx: transaction2,
+      scriptPubkey: 'pubkeyout2'
+    })
```
samholmes, line 180: I'm starting to understand what `ourIns` and `ourOuts` is. However, I'm a bit confused about why we track input/output ownership this way. You can see how you have two different transaction objects with identical inputs and outputs which is invalid from a consensus point of view. I'm aware this is dummy txs for testing, however shouldn't the processors check for conflicting txs like this?

Even if we used the the same transaction with a different `scriptPubkey` for each `insertTransaction` call, wouldn't we still be adding the same transaction into our data layer twice or would the last insert stomp on the previous? Would it make sense to let `scriptPubkey` be `scriptPubkeys: string[]` and allow for one or more; this way we can declare ownership of more than one `output` in the transaction (which is very possible for a transaction).

I may be misunderstanding things here, so I'm happy to discuss more.

samholmes (changes requested):
Let's discuss the feedback already given, and do a review together to make sure that we get a better peer-review/coverage of this work. Thanks!

src/common/utxobased/db/newProcessor.ts
```diff
@@ -161,26 +194,160 @@ export async function makeNewProcessor(
...
+            if (!transaction.ourOuts.includes(output.n.toString())) {
+              transaction.ourOuts.push(output.n.toString())
+            }
+          }
+        }
+
+        transaction.ourAmount = calculateTxAmount(transaction)
+
+        // Save tx by blockheight
+        const txidsByTransactionBlockHeight = await (
```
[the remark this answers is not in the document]
TheCharlatan: Done.

src/common/utxobased/db/newProcessor.ts
```diff
@@ -161,26 +194,160 @@ export async function makeNewProcessor(
...
-    async insertTransaction(_args: IProcessorTransaction): Promise<void> {
-      return
+    async insertTransaction(args: InsertTransactionArgs): Promise<void> {
+      const { scriptPubkey, tx } = args
+      return await baselets.tx(async tables => {
+        // Check if the transaction already exists
+        let processorTx: IProcessorTransaction | undefined
+        try {
+          processorTx = (await tables.txById.query('', [tx.txid]))[0]
+        } catch {}
```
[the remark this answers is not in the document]
TheCharlatan: I like it. Also any `let` that we can replace with a `const` is a win. Changed.

src/common/utxobased/db/newProcessor.ts
```diff
@@ -161,26 +194,160 @@ export async function makeNewProcessor(
...
+              [txId]
+            )
+            txs.push(...txById)
+          } catch {}
+        }
+        // Fetch transactions by min block height
+        if (blockHeightMax != null && blockHeight == null) blockHeight = 0
+        if (blockHeight != null) {
+          try {
+            const txidsByMinBlockHeight = await (
```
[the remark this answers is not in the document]
TheCharlatan: Thank you. Changed.

src/common/utxobased/db/newProcessor.ts
```diff
@@ -161,26 +194,160 @@ export async function makeNewProcessor(
...
 
     async fetchTransactions(
-      _args: FetchTransactionArgs
+      args: FetchTransactionArgs
     ): Promise<IProcessorTransaction[]> {
-      return []
+      const txs: IProcessorTransaction[] = []
+      await baselets.tx(async tables => {
+        const { blockHeightMax, txId, options } = args
+        let { blockHeight } = args
```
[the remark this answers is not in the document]
TheCharlatan: Done.

src/common/utxobased/db/newProcessor.ts
```diff
@@ -161,26 +194,160 @@ export async function makeNewProcessor(
...
+            )
+              .reverse()
+              .map(({ [RANGE_ID_KEY]: id }) => id)
+
+            const txsById: IProcessorTransaction[] = await tables.txById.query(
+              '',
+              txidsByMinBlockHeight
+            )
+            txs.push(...txsById)
+          } catch {}
```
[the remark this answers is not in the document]
TheCharlatan: Thanks for bringing this up. This was also not properly done in the current/deprecated processor. There we sometimes check the exception message:
```
  try {
    // Create index for size of value
    await tables.utxoIdsBySize.insert('', ...)
  } catch (err) {
    if (err.message !== 'Cannot insert data because id already exists') {
      throw err
    }
```
And sometimes we just ignored errors, like error. It's also why the error handling in the tests is wonky to say the least. There should be a middle ground between not throwing and throwing only when something is really wrong, and it's not just that we don't have any results for a certain key.  I'll research some more how we could go about typing the errors..

test/common/utxobased/db/NewProcessor.spec.ts
TheCharlatan: Yeah, fair point. I'll add arguments where needed.

test/common/utxobased/db/NewProcessor.spec.ts
TheCharlatan: It does convey that they may not be related, and that we are passing in the matching script pubkey 'by chance' . But at the same time, I agree with you, since we are testing this very case (e.g. scriptPubkey matches with the one in output 1). Changed it to match your suggestion and included references to the txid's as well where applicable.

test/common/utxobased/db/NewProcessor.spec.ts
TheCharlatan: Yes, it does make somewhat sense from a network perspective. We have network calls, where the network does not know for sure if you already have the transaction, and we have some where we can be certain that the transaction is already saved.

test/common/utxobased/db/NewProcessor.spec.ts
TheCharlatan: > You can see how you have two different transaction objects with identical inputs and outputs which is invalid from a consensus point of view. I'm aware this is dummy txs for testing, however shouldn't the processors check for conflicting txs like this?

It's perfectly valid from a consensus point of view that transactions can have identical outputs. The input in our serialization here also contains the previous (now consumed output's) transaction id. This txid needn't be unique either! See this stackexchange question: https://bitcoin.stackexchange.com/questions/75300/what-would-happened-if-two-transactions-have-the-same-hash . Check how the coinbase transactions here: https://www.blockchain.com/btc/block/91812 and here: https://www.blockchain.com/btc/block/91842 have the same txid! Nowadays it's super uncommon for this to happen, since there are policy rules in place making it harder, but it's not cryptographically impossible.

> Even if we used the the same transaction with a different scriptPubkey for each insertTransaction call, wouldn't we still be adding the same transaction into our data layer twice or would the last insert stomp on the previous? 

The last insert stomps the previous, updating ourIns and ourOuts based on the passed in scriptPubkey.

> Would it make sense to let scriptPubkey be scriptPubkeys: string[] and allow for one or more; this way we can declare ownership of more than one output in the transaction (which is very possible for a transaction).

Not really. The network layer passes in one or more transactions for a single scriptPubkey at a time. This means for a transaction actually containing one ourIn and one ourOut, we insert it twice. The first time with the scriptPubkey of the input populating ourIn, the second time with the scriptPubkey of the output populating ourOut.

samholmes (approved):

PR #170, opened 2021-08-26
NewProcessor: Add UTXO processing

samholmes:


src/common/utxobased/db/newProcessor.ts
```diff
@@ -151,20 +151,30 @@ export async function makeNewProcessor(
...
+    async updateUtxo(utxo: IUTXO): Promise<void> {
+      return await baselets.utxo(async tables => {
+        await tables.utxoById.insert('', utxo.id, utxo)
+      })
     },
 
-    async fetchUtxos(_utxoIds: string[]): Promise<IUTXO[]> {
-      return []
+    async fetchUtxos(utxoIds: string[]): Promise<IUTXO[]> {
+      if (utxoIds.length === 0) return []
```
TheCharlatan, line 173: I just noticed, that in this case, we should be returning all UTXO's. It's also already documented in the interface declaration: `  // fetch either all UTXOs if the array is empty or as selected from an array
  // of UTXO ids`

samholmes: I had to use `dumpData` to do this because `query` for `HashBase` wont allow an empty array of hashes.

TheCharlatan (approved):

PR #171, opened 2021-08-29
New processor dump data

TheCharlatan:


src/common/utxobased/db/newProcessor.ts
```diff
@@ -53,8 +53,14 @@ interface BlockHeightArgs {
   thresholdBlocks: number
 }
 
+interface DumpDataReturn {
+  databaseName: string
+  data: never
```
samholmes, line 58: Why is the type `never`?

src/common/utxobased/db/newProcessor.ts
```diff
@@ -151,6 +157,16 @@ export async function makeNewProcessor(
...
 
+    async dumpData(): Promise<DumpDataReturn[]> {
+      const allBases = Object.values(baselets.all)
+      return await Promise.all(
+        allBases.map(async base => ({
+          databaseName: base.databaseName,
+          data: (await base.dumpData('')) as never
+        }))
+      )
+    },
```
samholmes, line 168: What is the reason for `dumpData()` method? I'm assuming it'll be needed in the engine, but for what or where?

samholmes (changes requested):
I'm not sure if `never` is the correct type for the data, so perhaps we should reconsider.

src/common/utxobased/db/newProcessor.ts
samholmes: Use type `unknown`.

samholmes (approved):
Approved after changes

samholmes (approved):

samholmes:
Added some fixups to another branch here: https://github.com/EdgeApp/edge-currency-plugins/tree/newProcessorIntegration

PR #177, opened 2021-09-08
Fix lastUsed function signature

TheCharlatan:
Also ended up fixing an off-by-one error in the setLookAhead while loop.
    
The initial conditions and last valid iteration, around the edges
of the control logic in the while loop with gap limit 20 are now:

```
lastUsed: -1 , addressCount: 0 => -1 + 20 >= 19
lastUsed: 0, addressCount: 0 => 0 + 20 >= 20
```

### DIFF
--- a/src/common/utxobased/engine/makeUtxoEngineState.ts
+++ b/src/common/utxobased/engine/makeUtxoEngineState.ts
@@ -400,7 +400,7 @@ const setLookAhead = async (args: SetLookAheadArgs): Promise<void> => {
     }
 
     const getLastUsed = async (): Promise<number> =>
-      await processor.lastUsedIndexByFormatPath({ ...partialPath })
+      (await processor.lastUsedIndexByFormatPath({ ...partialPath })) ?? -1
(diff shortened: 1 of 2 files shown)


src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -399,12 +399,12 @@ const setLookAhead = async (args: SetLookAheadArgs): Promise<void> => {
       changeIndex: branch
     }
 
-    const getLastUsed = async (): Promise<number> =>
+    const getLastUsed = async (): Promise<number | undefined> =>
```
samholmes, line 402: Include the `?? -1` within `getLastUsed` and keep the return type as `Promise<number>`.

samholmes (approved):
Makes sense. Good catch. Minor fixup suggestion.

src/common/utxobased/engine/makeUtxoEngineState.ts
TheCharlatan: Fixed.

PR #179, opened 2021-09-10
Add mutex to processor utxo, address and tx groups

TheCharlatan:


samholmes (approved):
This will work for our purpose.

I almost expect a more generic table locking mechism similar to SQL's `BEGIN TRANSACTION`. Perhaps something similar to this could be provided by baselet in the future:

```ts
transaction<T extends Baselet[]>(baslets: T, fn: (baselets: T) => void): Promise<void>
```

For implementation, somehow `transaction` will lock all baselets using `async-mutex` and release them back once the transaction `fn` resolves.

PR #181, opened 2021-09-13
Upgrade baselet

samholmes:
This upgrades baselet to the new branch which includes type safety on all types of baselets. Includes necessary changes to the processor and engine with the added type safety/coverage.

### DIFF
--- a/src/common/utxobased/db/makeBaselets.ts
+++ b/src/common/utxobased/db/makeBaselets.ts
@@ -1,85 +1,66 @@
 import { Mutex } from 'async-mutex'
 import {
-  BaseType,
-  createCountBase,
-  createHashBase,
-  createRangeBase,
-  openBase
+  createOrOpenCountBase,
+  createOrOpenHashBase,
+  createOrOpenRangeBase
 } from 'baselet'
-import { CountBase } from 'baselet/src/CountBase'
-import { HashBase } from 'baselet/src/HashBase'
-import { RangeBase } from 'baselet/src/RangeBase'
 import { Disklet } from 'disklet'
 
 import {
-  addressByScriptPubkeyConfig,
-  addressPathByMRUConfig,
-  blockHashByBlockHeightConfig,
-  lastUsedByFormatPathConfig,
-  RANGE_ID_KEY,
-  RANGE_KEY,
-  scriptPubkeyByPathConfig,
-  scriptPubkeysByBalanceConfig,
-  spentUtxoByIdConfig,
-  txByIdConfig,
-  txIdsByBlockHeightConfig,
-  txIdsByDateConfig,
-  txsByDateConfig,
-  txsByScriptPubkeyConfig,
-  usedFlagByScriptPubkeyConfig,
-  utxoByIdConfig,
-  utxoIdsByScriptPubkeyConfig,
-  utxoIdsBySizeConfig
+  AddressByScriptPubkeyBaselet,
+  addressByScriptPubkeyOptions,
+  BlockHashByBlockHeightBaselet,
+  blockHashByBlockHeightOptions,
+  LastUsedByFormatPathBaselet,
+  lastUsedByFormatPathOptions,
+  ScriptPubkeyByPathBaselet,
+  scriptPubkeyByPathOptions,
+  TxByIdBaselet,
+  txByIdOptions,
+  TxIdByBlockHeightBaselet,
+  TxIdByDateBaselet,
+  txIdsByBlockHeightOptions,
+  txIdsByDateOptions,
+  UtxoByIdBaselet,
+  utxoByIdOptions,
+  UtxoIdsByScriptPubkeyBaselet,
+  utxoIdsByScriptPubkeyOptions
 } from './Models/baselet'
-import { Baselet, BaseletConfig } from './types'
-
-async function createOrOpen(
-  disklet: Disklet,
-  config: BaseletConfig<BaseType.HashBase>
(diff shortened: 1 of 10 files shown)


src/common/utxobased/db/makeBaselets.ts
```diff
@@ -102,37 +117,37 @@ export interface Databases {
...
-  txIdsByDate: RangeBase
+  txsByDate: RangeBase<IProcessorTransaction>
 }
 
 export interface UTXOTables {
-  utxoById: HashBase
+  utxoById: HashBase<IUTXO>
   // deprecated
-  utxoIdsByScriptPubkey: HashBase
+  utxoIdsByScriptPubkey: HashBase<string>
```
samholmes, line 144: Type should be `HashBase<string[]>`. This will incur a necessary fix in the implemenatation.

src/common/utxobased/db/makeProcessor.ts
```diff
@@ -205,22 +205,24 @@ export async function makeProcessor(
...
       const { scriptPubkey, utxoIds = [] } = args
       return await baselets.utxo(async tables => {
         if (scriptPubkey != null) {
-          const utxoIdsByScriptPubkey = (
+          const utxoIdsByScriptPubkey: string[] = (
             await tables.utxoIdsByScriptPubkey.query('', [scriptPubkey])
-          ).filter(utxoId => utxoId != null)
+          ).filter(utxoId => utxoId != null) as string[]
 
           utxoIds.push(...utxoIdsByScriptPubkey)
```
samholmes, line 228: Looks like we have a bug here. Once the type for `utxoIdsByScriptPubkey` has change to the correct type, we must make sure to change `saveUtxo` to first query from `utxoIdsByScriptPubkey`, modify the resultset, then save it back, then second, need to change the query here in `fetchUtxos` to spread on the first element of the resultset.

src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -1049,6 +1050,7 @@ const processUtxoTransactions = async (
   const currentUtxoIds: string[] = []
   let oldBalance = '0'
   for (const utxo of currentUtxos) {
+    if (utxo == null) continue
```
samholmes, line 1053: Throw on this assertion.

samholmes (commented):
Changes

src/common/utxobased/db/makeProcessor.ts
```diff
@@ -193,25 +193,49 @@ export async function makeProcessor(
...
+        }
+        for (const scriptPubkey of Object.keys(utxoIdsMap)) {
+          const utxoIdsToRemove = utxoIdsMap[scriptPubkey]
+          const [utxoIds = []] = await tables.utxoIdsByScriptPubkey.query('', [
+            scriptPubkey
+          ])
+          for (const utxoIdToRemove of utxoIdsToRemove) {
+            utxoIds.splice(utxoIds.indexOf(utxoIdToRemove), 1)
+          }
+          await tables.utxoIdsByScriptPubkey.insert('', scriptPubkey, utxoIds)
```
TheCharlatan, line 228: Is this missing the case where no utxoIds remain? Shouldn't this be something like:
```
if (Array.length(utxoIds) > 0)
    await tables.utxoIdsByScriptPubkey.insert('', scriptPubkey, utxoIds)
else 
    await tables.utxoIdsByScriptPubkey.delete(''', scriptPubkey)
```

TheCharlatan (changes requested):

src/common/utxobased/db/makeProcessor.ts
samholmes: I'm assuming you mean to say that we should delete the database entry once the `utxoIds` in the database becoomes empty. This is only a bit of extra cleanup on the hard-disk. After giving it some thought, I'm for this improvement.

src/common/utxobased/db/makeBaselets.ts
```diff
@@ -102,37 +117,37 @@ export interface Databases {
...
   // deprecated
   utxoIdsBySize: RangeBase
 }
 
 export interface BlockTables {
-  blockHashByBlockHeight: HashBase
+  blockHashByBlockHeight: HashBase<string>
 }
 
 // deprecated
```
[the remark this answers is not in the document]
TheCharlatan: Done in cc84dc5135b4529cfdfb4bace0909a89f7cc40a0

src/common/utxobased/db/makeBaselets.ts
```diff
@@ -102,37 +117,37 @@ export interface Databases {
...
   // deprecated
   usedFlagByScriptPubkey: HashBase
-  lastUsedByFormatPath: HashBase
 }
 
 export interface TransactionTables {
-  txById: HashBase
+  txById: HashBase<IProcessorTransaction>
+  txIdsByBlockHeight: RangeBase<TxIdsByBlockHeight>
+  txIdsByDate: RangeBase<TxIdByDate, 'date', 'txId'>
```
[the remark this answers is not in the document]
samholmes: Decided against this. The way to solve for simplicity is not to have a second "index type" which is passed around, but rather to just pass around the "database type" (RangeBase type). This means you define the RangeBase type once and use that in the implementation to create the instance; with the instance you have access to the key names via added properties on the RangeBase type.

samholmes:
Replaced by: https://github.com/EdgeApp/edge-currency-plugins/pull/189

PR #158, opened 2021-06-23
Fixup balance calculation

TheCharlatan:

PR #182, opened 2021-09-15
Fixup balance calculation

samholmes:
- Stop using Set's in the engine
- Make sure that objects are only mutated atomically

_Replaces https://github.com/EdgeApp/edge-currency-plugins/pull/158_

My only concern with changing the `utxos` field from a Set to an array is one about performance. Our algorithm goes from an O(1) to O(n) if I'm not mistaken. How large is this dataset; if large then there is a concern, if not then no need to optimize.


samholmes:
This was meant to be dropped. It doesn't fix anything.

PR #184, opened 2021-09-18
Grab blockHeight before address query and save height query results

TheCharlatan:


samholmes:
Replaced by: https://github.com/EdgeApp/edge-currency-plugins/pull/190

PR #190, opened 2021-09-23
Grab blockHeight before address query and save height query results

samholmes:
Replaces: https://github.com/EdgeApp/edge-currency-plugins/pull/184

### DIFF
--- a/src/common/utxobased/engine/makeUtxoEngineState.ts
+++ b/src/common/utxobased/engine/makeUtxoEngineState.ts
@@ -207,6 +207,7 @@ export function makeUtxoEngineState(
           response.address,
           path.format,
           path.branch,
+          0,
           taskCache.transactionsCache
         ).catch(() => {
           throw new Error('failed to add to transaction cache')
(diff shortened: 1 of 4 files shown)


src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -464,22 +463,23 @@ const addToTransactionCache = async (
...
   const { walletTools, processor } = args
-  // Fetch the networkQueryVal from the database
+  // Fetch the blockHeight for the address from the database
   const scriptPubkey = walletTools.addressToScriptPubkey(address)
-  const { networkQueryVal = 0 } =
-    (await processor.fetchAddress(scriptPubkey)) ?? {}
+
   if (blockHeight === 0) {
-    blockHeight = networkQueryVal
+    const { blockHeight: addressBlockHeight = 0 } =
```
TheCharlatan, line 470: I don't think this change is correct. The logic flow before was:

1. Add a derived address to  the TransactionCache based on its own networkQueryVal (which initially is zero)
2. Add the current blockheight to the cache
3. Query the transactions based on the networkQueryVal (0)
4. Once the query is successful, update the address' networkQueryVal of the address to the blockHeight

Now we appear to have:
1. Add a derived address to the TransactionCache based on the blockHeight (the current height)
2. Query the transactions based on the blockHeight (current height)
3. Once the query is successful, update the address' blockHeight with the current height

The rename should keep two different names in my opinion. How about `lastQueriedBlockHeight` and `currentBlockHeight`?

TheCharlatan (approved):

PR #183, opened 2021-09-15
Non-processor related cleaners

TheCharlatan:
Replaces https://github.com/EdgeApp/edge-currency-plugins/pull/151

PR #191, opened 2021-09-23
Non-processor related cleaners

samholmes:
Replaces: https://github.com/EdgeApp/edge-currency-plugins/pull/183


samholmes:
@TheCharlatan We should discuss what you mean by:

> The response for an address message is a mix type, that could not be
> easily implemented by the cleaner. Further investigation on its
> implementation is required.

In this commit: https://github.com/EdgeApp/edge-currency-plugins/pull/191/commits/37ca3d4d0262e5dd18a400c2b8755184a2362c6c (see the commit message).



cc @thehobbit85

src/common/utxobased/network/Socket.ts
```diff
@@ -36,7 +39,8 @@ export interface Socket {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   submitTask: (task: WsTask<any>) => void
   onQueueSpace: (cb: OnQueueSpaceCB) => void
-  subscribe: (subscription: WsSubscription) => void
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  subscribe: (subscription: WsSubscription<any>) => void
```
samholmes, line 43: Use `<T>` generic.

src/common/utxobased/network/Socket.ts
```diff
@@ -60,7 +64,7 @@ interface WsMessage {
 }
 
 interface Subscriptions {
-  [key: string]: WsSubscription
+  [key: string]: WsSubscription<unknown>
```
samholmes, line 67: Make it generic; `Subscriptions<T>`

src/common/utxobased/network/Socket.ts
```diff
@@ -171,7 +175,9 @@ export function makeSocket(uri: string, config: SocketConfig): Socket {
     }
   }
 
-  const subscribe = (subscription: WsSubscription): void => {
+  // add any exception, since the passed in template parameter needs to be re-assigned
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const subscribe = (subscription: WsSubscription<any>): void => {
```
samholmes, line 180: Change to generic.

src/common/utxobased/network/Socket.ts
```diff
@@ -271,7 +277,11 @@ export function makeSocket(uri: string, config: SocketConfig): Socket {
             if (!subscription.subscribed) {
               subscription.deferred.reject()
             }
-            subscription.cb(json.data)
+            if (subscription.cleaner != null) {
+              subscription.cb(subscription.cleaner(json.data))
+            } else {
+              subscription.cb(json.data)
+            }
```
samholmes, line 284: No need for null case check because cleaner is not optional on the subscription object type.

src/common/utxobased/network/Socket.ts
```diff
@@ -348,7 +363,7 @@ export function makeSocket(uri: string, config: SocketConfig): Socket {
       onQueueSpace = cb
     },
 
-    subscribe(subscription: WsSubscription): void {
+    subscribe(subscription: WsSubscription<unknown>): void {
```
samholmes, line 366: Generic.

src/common/plugin/pluginState.ts
```diff
@@ -1,11 +1,12 @@
...
 import { navigateDisklet } from 'disklet'
 import { EdgeIo, EdgeLog } from 'edge-core-js/types'
 import { makeMemlet, Memlet } from 'memlet'
 
 import { UtxoEngineState } from '../utxobased/engine/makeUtxoEngineState'
-import { ServerCache, ServerInfo } from './serverCache'
+import { asServerInfo, ServerCache, ServerInfo } from './serverCache'
 
 const InfoServer = 'https://info1.edge.app/v1'
 const FixCurrencyCode = (currencyCode: string): string => {
```
samholmes, line 12: Remove `FixCurrencyCode` because we're not using electrum servers anymore.

src/common/plugin/pluginState.ts
```diff
@@ -109,7 +110,9 @@ export class PluginState extends ServerCache {
 
   async load(): Promise<PluginState> {
     try {
-      this.serverCacheJson = await this.memlet.getJson('serverCache.json')
+      const serverCacheText = await this.memlet.getJson('serverCache.json')
+      const cleaner = asObject(asServerInfo)
+      this.serverCacheJson = cleaner(JSON.parse(serverCacheText))
```
samholmes, line 115: Add cleaners to `fetchServers`. That's where we need to use cleaners.

src/common/plugin/pluginState.ts
samholmes: Either we do all the work for the server-list task in this PR or we do the server-list work in it's own PR and rebase this PR onto that one.

src/common/utxobased/network/BlockBookAPI.ts
```diff
@@ -27,20 +35,81 @@ export const infoMessage = (): PartialTask => {
   }
 }
 
+export const asIServerInfo = asObject({
```
thehobbit85, line 38: Say, what's with this "I" everywhere? as in "asIServerInfo", since when are we apple? ;)

src/common/utxobased/network/Socket.ts
```diff
@@ -54,27 +55,29 @@ interface SocketConfig {
...
 }
 
 export function makeSocket(uri: string, config: SocketConfig): Socket {
   let socket: InnerSocket | null
   const { emitter, log, queueSize = 50, walletId } = config
   log('makeSocket connects to', uri)
   const version = ''
-  const subscriptions: Subscriptions = {}
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const subscriptions: Subscriptions<any> = {}
```
thehobbit85, line 77: Why not used "mixed" or "unknown" instead of any? Would that work?

src/common/utxobased/network/Socket.ts
```diff
@@ -54,27 +55,29 @@ interface SocketConfig {
...
   let onQueueSpace = config.onQueueSpaceCB
-  let pendingMessages: PendingMessages = {}
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let pendingMessages: PendingMessages<any> = {}
```
thehobbit85, line 80: Same as the comment above

src/common/plugin/makeMetadata.ts
```diff
@@ -76,7 +76,7 @@ export const makeMetadata = async (
 const fetchMetadata = async (memlet: Memlet): Promise<LocalWalletMetadata> => {
   try {
     const dataStr = await memlet.getJson(metadataPath)
-    return JSON.parse(dataStr)
+    return asLocalWalletMetadata(JSON.parse(dataStr))
```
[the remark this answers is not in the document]
samholmes: Nice catch!

src/common/fees/makeFees.ts
```diff
@@ -108,11 +115,13 @@ export const makeFees = async (config: MakeFeesConfig): Promise<Fees> => {
 const fetchCachedFees = async (
   memlet: Memlet,
   currencyInfo: EngineCurrencyInfo
```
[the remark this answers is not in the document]
samholmes: Seems like the parameter should have simply been a fallback. Switched it to asking for a fallback

src/common/utxobased/network/BlockBookAPI.ts
samholmes: I completely agree. I have some branches which will rename these identifies to do away the meaningless I prefix. I see it as a "c-sharpy thing"; Microsoft-y.

src/common/utxobased/network/Socket.ts
samholmes: `mixed` is a Flow thing. `unknown` would constrain `subscriptions`, we want subscriptions to be unconstrained so it can include any WsSubscription. `subscriptions` is an implementation detail, so it's _slightly_ acceptable if we don't have strict type safety within the implementation?

PR #192, opened 2021-09-24
Bch replay protection

TheCharlatan:


src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -241,31 +243,66 @@ export async function makeUtxoEngine(
...
+      const totalAmountToSend = spendTargets.reduce(
+        (sum, { nativeAmount }) => bs.add(sum, nativeAmount ?? '0'),
+        '0'
+      )
+      const utxos =
+        options?.utxos ?? (await processor.fetchUtxos({ utxoIds: [] }))
+
+      if (bs.gt(totalAmountToSend, `${sumUtxos(utxos)}`)) {
+        throw new InsufficientFundsError(currencyInfo.currencyCode)
+      }
```
samholmes, line 260: This should be a separate PR fix; unrelated to this PR.

src/common/utxobased/engine/makeUtxoWalletTools.ts
```diff
@@ -1,9 +1,12 @@
 import { AddressPath, CurrencyFormat, NetworkEnum } from '../../plugin/types'
+import { cdsScriptTemplates } from '../keymanager/bitcoincashUtils/checkdatasig'
```
samholmes, line 2: Propagate the scriptTemplate in the currency's info (reference the existing bitcoin plugin architecture).

samholmes (changes requested):

src/common/utxobased/engine/makeUtxoWalletTools.ts
TheCharlatan: I pushed a fixup to implement this.

src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -881,7 +891,7 @@ const internalDeriveScriptAddress = async (
   return { address, scriptPubkey, redeemScript }
 }
 
-interface GetFreshAddressArgs extends FormatArgs { }
+interface GetFreshAddressArgs extends FormatArgs {}
```
thehobbit85, line 894: I know it's not part of the PR but on line 1234, why we are not using a const here:
``` const [tx] = await processor.fetchTransactions({ txId: utxo.txid })```

File: src/common/utxobased/info/scriptTemplates/types.ts [hunk 4]
```diff
@@ -0,0 +1,14 @@
+export interface ScriptTemplates {
+  [key: string]: ScriptTemplate
+}
+
+export type ScriptTemplate = (
+  pubkey: string,
+  optionalArgs?: OptionalScriptTemplateArgs
+) => string
+
+export interface OptionalScriptTemplateArgs {
```
thehobbit85, line 10: No need for this type here, try and keep the files outside of the info folder as "generic" as possible.

thehobbit85 (changes requested):
Minor changes, but overall, much better.

src/common/utxobased/info/scriptTemplates/types.ts
TheCharlatan: I could not come up with a better way to define these additional parameters, that does not involve either weakening the type, or a generic that would propagate the type selection up to the core.

src/common/utxobased/engine/makeUtxoEngineState.ts
TheCharlatan: https://github.com/EdgeApp/edge-currency-plugins/pull/198

PR #204, opened 2021-10-14
Server state fixes

samholmes:
This fixes a few issues with the server state:

1. Memlet's return value was incorrectly assumed to be a JSON string; remove the extra JSON parsing.
2. Add cleaners for the serverCache.json file data object.
3. Add cleaners for the info server network responses for the server lists.

### DIFF
--- a/src/common/plugin/pluginState.ts
+++ b/src/common/plugin/pluginState.ts
@@ -6,7 +6,7 @@ import { EdgeIo, EdgeLog } from 'edge-core-js/types'
 import { makeMemlet, Memlet } from 'memlet'
 
 import { UtxoEngineState } from '../utxobased/engine/makeUtxoEngineState'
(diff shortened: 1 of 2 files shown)


src/common/plugin/pluginState.ts
```diff
@@ -149,32 +162,32 @@ export class PluginState extends ServerCache {
...
-      const remoteServerList: string[] = serverListResponse[this.currencyCode]
+          return
+        }
+        return await response.json()
+      })
+      .catch(err => {
+        this.log(
+          `${this.pluginId} - Fetching ${serverListInfoUrl} failed: ${err.message}`
+        )
+      })
```
thehobbit85, line 185: Refactor this to use try/catch instead of this weird promise/async combo.

thehobbit85 (changes requested):

PR #207, opened 2021-10-15
Server Score / Plugin State refactor

TheCharlatan:
This removes the in-memory cache from the server state and re-implements the logic in the plugin state. The server state file is renamed to server scores. 

---
  - https://app.asana.com/0/0/1199373123392598


TheCharlatan:
Rebased.

samholmes:
Rebased onto master, tested it works, and added a couple of cleanup commits.

samholmes:
Removed one of my last commits after an OTS review with @thehobbit85. I can say that this is now ready for merge.

PR #209, opened 2021-10-15
Refactor EngineCurrencyInfo

samholmes:
Refactor EngineCurrencyInfo for each wallet to no longer extend EdgeCurrencyInfo. Rather, have EngineInfo and EdgeCurrencyInfo be sibling fields in a PluginInfo object which is passed to makeCurrencyPlugin.


samholmes (commented):
Approved upon fixing CI if the tests were broken by this PR.

PR #210, opened 2021-10-18
Package cleanup

samholmes:
Adds rollup and "modernizes" the package's scripts.


test/common/plugin/currencyPlugins.spec.ts
```diff
@@ -12,7 +12,7 @@ import {
 } from 'edge-core-js/types'
 
 import fixtures from '../../../fixtures/currencyPlugins.json'
-import plugin from '../../../src/btc'
+import plugin from '../../../src'
```
swansontec, line 15: Still needs to be `/index`. The Node behavior of auto-guessing "index" is non-standard.

yarn.lock
```diff
@@ -7062,6 +7173,18 @@ sucrase@^3.12.1:
...
+  resolved "https://registry.yarnpkg.com/sucrase/-/sucrase-3.20.2.tgz#28a28dc58a55be0d6916d5c9b2440d203e9ffe62"
+  integrity sha512-EdJ5M6VEvToIZwIWiZ71cxe4CklDRG8PdSjUSst+BZCUGlaEhnrdQo/LOXsuq3MjWRbfepg1XTffClK0Tmo0HQ==
+  dependencies:
+    commander "^4.0.0"
+    glob "7.1.6"
+    lines-and-columns "^1.1.6"
+    mz "^2.7.0"
+    pirates "^4.0.1"
+    ts-interface-checker "^0.1.9"
+
```
swansontec, line 7187: Please delete both `sucrase` blocks & re-run yarn to consolidate this dependency.

rollup.config.js
```diff
@@ -0,0 +1,36 @@
+import resolve from '@rollup/plugin-node-resolve'
+import sucrase from '@rollup/plugin-sucrase'
+import mjs from 'rollup-plugin-mjs-entry'
+
+import packageJson from './package.json'
+
+// Matches all dependencies and the nested modules of the dependencies
+const dependencies = Object.keys(packageJson.dependencies).map(
+  dep => new RegExp(`^${dep}/?.*$`)
```
swansontec, line 9: The regexp for dependency "foo" would match "foobar". It would need to have the whole `/.*` part be optional, like `(/.*)?`

rollup.config.js
```diff
@@ -0,0 +1,36 @@
...
+      {
+        file: cjsFile,
+        format: 'cjs'
+      }
+    ],
+    plugins: [resolve(resolveOpts), sucrase(sucraseOpts), mjs()]
+  }
+})
+
+export default [...entryConfigs]
```
swansontec, line 36: Why can't we do `export default entryFiles.map(...`

package.json
```diff
@@ -6,6 +6,15 @@
...
   "license": "MIT",
   "author": "Airbitz, Inc.",
+  "exports": {
+    ".": {
+      "import": "./lib/index.mjs",
+      "require": "./lib/index.js"
+    }
+  },
+  "main": "./lib/index.js",
+  "module": "./lib/index.mjs",
```
swansontec, line 16: This is invalid. You cannot use the mjs file as the "module" entry point, since the mjs file is just a wrapper around the CJS file. The "module" entry is supposed to be a native ESM file, to allow bundlers like Webpack to do special optimizations not possible using CommonJS.

package.json
```diff
@@ -79,31 +105,5 @@
...
-    "cleaners": "^0.3.12",
-    "disklet": "^0.4.5",
-    "edge-core-js": "^0.18.9",
-    "memlet": "^0.0.4",
-    "uri-js": "^4.4.0",
-    "url-parse": "^1.5.2",
-    "wif-smart": "^2.0.0",
-    "wifgrs": "^2.0.6",
-    "ws": "^7.3.1"
-  },
```
swansontec: Please sort package.json as the first commit in this series.

swansontec (changes requested):
There are some issues with the rollup & package.json exports setup.

swansontec (commented):
Much improve. Wow. Such package.

PR #217, opened 2021-10-21
Fix sync issues

samholmes:
Fixes two off-by-one errors in `setLookAhead` and an edge case failure in the condition for initial address subscriptions. The off-by-one errors fix thrown errors which freezes syncs. And the condition change fixes silent frozen syncs.

### DIFF
--- a/src/common/utxobased/engine/makeUtxoEngineState.ts
+++ b/src/common/utxobased/engine/makeUtxoEngineState.ts
@@ -415,8 +415,13 @@ const setLookAhead = async (
     })
 
     // Initialize the addressSubscribeCache with the existing addresses already
-    // processed by the processor. This happens only once; when the cache is empty.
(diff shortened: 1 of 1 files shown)


src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -118,11 +118,7 @@ export function makeUtxoEngineState(
   let processedPercent = 0
   const onAddressChecked = async (): Promise<void> => {
     processedCount = processedCount + 1
-    const totalCount = await getTotalAddressCount({
-      walletInfo,
-      currencyInfo,
-      processor
-    })
+    const totalCount = await getTotalAddressCount(walletInfo, processor)
```
thehobbit85, line 121: if (totalCount < currencyInfo.gapLimit) {
totalCount = currencyInfo.gapLimit
}

thehobbit85 (changes requested):
Address the comment

PR #219, opened 2021-10-22
Keymanager: Add support for legacy digibyte WIF

TheCharlatan:


src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -846,11 +846,31 @@ export function wifToPrivateKey(args: WIFToPrivateKeyArgs): string {
...
+        legacy: true
+      })
+      privateKey = bitcoin.ECPair.fromWIF(
+        args.wifKey,
+        network,
+        coinClass.bs58DecodeFunc
+      ).privateKey
+    } else {
+      throw err
+    }
```
samholmes, line 872: Hmm, my only hesitation for this implementation is that we assume the error to be an invalid WIF error and try with a legacy WIF. This means that we might enter the legacy WIF case incorrectly if the error is not from the WIF being invalid. Perhaps an error check would do, but better yet...Can't you pass an array of `BitcoinJSNetwork` objects as the second argument to `fromWIF` and omit the try..catch altogether?

samholmes (changes requested):

src/common/utxobased/keymanager/keymanager.ts
TheCharlatan: > Hmm, my only hesitation for this implementation is that we assume the error to be an invalid WIF error and try with a legacy WIF. This means that we might enter the legacy WIF case incorrectly if the error is not from the WIF being invalid

Yes, that is what I do now the PR replacing this one: https://github.com/EdgeApp/edge-currency-plugins/pull/235

> Perhaps an error check would do, but better yet...Can't you pass an array of BitcoinJSNetwork objects as the second argument to fromWIF and omit the try..catch altogether?

In this case we need to change the library, which I would like to avoid to keep it compatible with other changes from the library's upstream.

TheCharlatan:
Replacing this with: https://github.com/EdgeApp/edge-currency-plugins/pull/235

PR #224, opened 2021-10-26
Fix disableFetchingServers

samholmes:
The problem addressed with this commit is that the
`disableFetchingServers` currency info flag would halt the `fetchServer`
method completely.
The `fetchServer` method was also responsible for saving the server
list whether that's the default list or from the info server.

The solution is to refactor the parts in `fetchServer` routine which
saves the server list to a new `refreshServers` method for better
separation of concerns.


thehobbit85 (approved):

thehobbit85 (approved):
Don't merge it yet, I want to discuss something first.

PR #228, opened 2021-11-01
Remove rollup

samholmes:
During discussion about the known issue with production builds (bundles) caused by introducing rollup into the build steps for this project, it was decided we move away from bundling this project and leave bundling up to the app as a best practice. This PR reflects this direction.

_PR is a draft because I want to run this by @swansontec for a alternative opinion_


samholmes:
I've determined that the alternative to removing rollup is to live with potential uncaught issues during development that only show up in the production builds and hopefully caught by QA. 

The fix with the module import is a somewhat mysterious issue considering that the compiled JS production output from webpack/babel appears to be incorrect in how the imported `parse` method is invoked; it does appear to be undefined because of an incorrect import, but rather undefined because of a variable `t` is reused for unshifting an element within the scope where it should be the module... Perhaps this is a bug within webpack or babel. Overall it isn't worth the engineering effort to resolve at this time.

PR #230, opened 2021-11-03
Sam/Fix/Numb Keys

samholmes:
Re-implement `derivePublicKey` to follow `edge-core-js` key-formats spec

- Add PrivateKey and PublicKey types and asPrivateKey and asPublicKey
cleaners for internal use and documentation.
- Desensitize public keys returned by derivePublicKey using the cleaners.
- Modify util functions to always include segwit and wrapped-segwit
formats together, following the spec.

---
  - https://app.asana.com/0/0/1201256727099457

### DIFF
--- a/src/common/plugin/makeCurrencyTools.ts
+++ b/src/common/plugin/makeCurrencyTools.ts
@@ -1,5 +1,6 @@
 import * as bn from 'biggystring'
 import * as bip39 from 'bip39'
+import { uncleaner } from 'cleaners'
 import {
   EdgeCurrencyTools,
   EdgeEncodeUri,
@@ -13,10 +14,13 @@ import {
 import * as uri from 'uri-js'
 import urlParse from 'url-parse'
 
-import * as utxoUtils from '../utxobased/engine/utils'
-import { CurrencyFormatKeys } from '../utxobased/engine/utils'
+import { parsePathname } from '../utxobased/engine/utils'
+import {
+  asNumbWalletInfo,
+  asPrivateKey,
+  PrivateKey
+} from '../utxobased/keymanager/cleaners'
 import { NetworkEnum, PluginInfo } from './types'
-import * as pluginUtils from './utils'
 import { getFormatsForNetwork } from './utils'
 
 /**
@@ -30,36 +34,29 @@ export function makeCurrencyTools(
 ): EdgeCurrencyTools {
   const { currencyInfo, engineInfo, coinInfo } = pluginInfo
 
+  const asCurrencyPrivateKey = asPrivateKey(coinInfo.name, coinInfo.coinType)
+  const wasCurrencyPrivateKey = uncleaner(asCurrencyPrivateKey)
+  const asCurrencyNumbWalletInfo = asNumbWalletInfo(pluginInfo)
+
   const fns: EdgeCurrencyTools = {
     async createPrivateKey(
-      walletType: string,
+      _walletType: string,
       opts?: JsonObject
     ): Promise<JsonObject> {
-      const mnemonicKey = pluginUtils.getMnemonicKey({
-        coin: coinInfo.name
-      })
       const mnemonic = bip39.entropyToMnemonic(Buffer.from(io.random(32)))
-      const keys: JsonObject = {
-        [mnemonicKey]: mnemonic
-      }
 
-      return {
-        ...keys,
+      const privateKey: PrivateKey = {
+        seed: mnemonic,
         format: opts?.format ?? engineInfo.formats?.[0] ?? 'bip44',
         coinType: opts?.coinType ?? coinInfo.coinType ?? 0
       }
+
+      return wasCurrencyPrivateKey(privateKey)
     },
 
     async derivePublicKey(walletInfo: EdgeWalletInfo): Promise<JsonObject> {
-      const key = utxoUtils.getXpubKey({ coin: coinInfo.name })
-      // TODO: which xpub should be saved? the root path (m) or hardened path with the wallet format path (m/{purpose}'/{coinType}'/{account}')?
-      const publicKey: CurrencyFormatKeys = utxoUtils.deriveXpubsFromKeys({
-        keys: walletInfo.keys,
-        coin: coinInfo.name,
-        network: engineInfo.networkType ?? NetworkEnum.Mainnet
-      })
-      walletInfo.keys[key] = publicKey
-      return walletInfo.keys
+      const numbWalletInfo = asCurrencyNumbWalletInfo(walletInfo)
+      return numbWalletInfo.keys.publicKey
     },
 
     async parseUri(uri: string): Promise<EdgeParsedUri> {
@@ -85,7 +82,7 @@ export function makeCurrencyTools(
       const parsedUri: EdgeParsedUri = {}
       // Parse the pathname and add it to the result object
       if (pathname !== '') {
-        const parsedPath = utxoUtils.parsePathname({
+        const parsedPath = parsePathname({
           pathname: uriObj.pathname,
           coin: coinInfo.name,
           network: engineInfo.networkType ?? NetworkEnum.Mainnet
--- a/src/common/plugin/types.ts
+++ b/src/common/plugin/types.ts
@@ -1,5 +1,5 @@
 import { bip32 } from 'altcoin-js'
-import { asNumber, asObject, asString } from 'cleaners'
+import { asNumber, asObject, asString, asValue } from 'cleaners'
 import { Disklet } from 'disklet'
 import {
   EdgeCurrencyEngineOptions,
@@ -20,7 +20,8 @@ export enum NetworkEnum {
   Testnet = 'testnet'
 }
 
-export type CurrencyFormat = 'bip32' | 'bip44' | 'bip49' | 'bip84'
+export type CurrencyFormat = ReturnType<typeof asCurrencyFormat>
+export const asCurrencyFormat = asValue('bip32', 'bip44', 'bip49', 'bip84')
 
 export interface AddressPath {
   format: CurrencyFormat
--- a/src/common/utxobased/engine/makeUtxoEngine.ts
+++ b/src/common/utxobased/engine/makeUtxoEngine.ts
@@ -1,4 +1,5 @@
 import * as bs from 'biggystring'
+import { asMaybe } from 'cleaners'
 import {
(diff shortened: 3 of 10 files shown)


src/common/plugin/makeCurrencyTools.ts
```diff
@@ -51,15 +51,18 @@ export function makeCurrencyTools(
     },
 
     async derivePublicKey(walletInfo: EdgeWalletInfo): Promise<JsonObject> {
-      const key = utxoUtils.getXpubKey({ coin: engineInfo.network })
-      // TODO: which xpub should be saved? the root path (m) or hardened path with the wallet format path (m/{purpose}'/{coinType}'/{account}')?
-      const publicKey: CurrencyFormatKeys = utxoUtils.deriveXpubsFromKeys({
+      const privateKey = asPrivateKey(
```
samholmes, line 54: Define an `asCurrencyPrivateKey` alias cleaner in the outerscope.

src/common/plugin/types.ts
```diff
@@ -17,7 +17,8 @@ export enum NetworkEnum {
   Testnet = 'testnet'
 }
 
-export type CurrencyFormat = 'bip32' | 'bip44' | 'bip49' | 'bip84'
+export type CurrencyFormat = ReturnType<typeof asCurrencyFormat>
+export const asCurrencyFormat = asValue('bip32', 'bip44', 'bip49', 'bip84')
```
samholmes, line 21: Discover if we can remove `bip84`.

samholmes (commented):
Remove all the utils cruft in favor of a simple object map that maps the private key formats to the public key formats based on the key-formats standard. The map is then reduced to the public key format. Don't use `if` statements; this PR should be -200 lines diff.

src/common/plugin/types.ts
samholmes: I don't think we can safely remove this from the union type because this type appears to be a general purpose union of all formats.

samholmes:
Feedback from discussions with @swansontec : 

* Remove `WalletKeys` and don't pass along `walletKeys` (which includes private keys) throughout the plugin. Instead only pass along the `PublicKey` and `PrivateKey` objects where needed. We should avoid passing along the sensitive data that is in either `walletInfo` and `walletKeys`.
* Change `PublicKey` type from a union of format records to a single record where the formats are optional.

src/common/utxobased/keymanager/cleaners.ts
```diff
@@ -0,0 +1,164 @@
...
+          'Cannot derive public key from watch-only wallet (no private key)'
+        )
+      }
+
+      const currencyFormatKeys = deriveXpubsFromKeys({
+        privateKey,
+        coin: coinName,
+        network: networkType ?? NetworkEnum.Mainnet
+      })
+      return asPublicKey({
```
swansontec, line 148: This data has a known data type, so it should not need to be cleaned. Please replace the `asPublicKey` with a type delaration on the IIFE return value, so Typescript checks it instead of the cleaner.

src/common/utxobased/engine/utils.ts
```diff
@@ -202,12 +165,12 @@ export const deriveXprivFromKeys = (args: {
...
 
 export const deriveXpubsFromKeys = (args: {
-  keys: UtxoKeyFormat
+  privateKey: PrivateKey
   coin: string
   network: NetworkEnum
 }): CurrencyFormatKeys => {
   const xpubs: CurrencyFormatKeys = {}
-  for (const format of getWalletSupportedFormats(args)) {
+  for (const format of asWalletKeysFormats(args.privateKey)) {
```
swansontec, line 173: `args.privateKey` is typed, and therefore should *not* be cleaned. Instead, do `getSupportedFormats(privateKey.format)`.

src/common/utxobased/keymanager/cleaners.ts
```diff
@@ -0,0 +1,164 @@
...
+    case 'bip44':
+      return ['bip44']
+    case 'bip49':
+      return ['bip49', 'bip84']
+    case 'bip84':
+      return ['bip49', 'bip84']
+    default:
+      throw new Error(`Unsupported format ${format}`)
+  }
+}
```
swansontec, line 99: This should not be a cleaner - it is being used in contexts where data has already been cleaned, so it should just be a normal utility function taking `CurrencyFormat` and returning `CurrencyFormat[]`. Let's call it `getSupportedFormats`, for the sake of this PR.

src/common/utxobased/keymanager/cleaners.ts
```diff
@@ -0,0 +1,164 @@
...
+      const currencyFormatKeys = deriveXpubsFromKeys({
+        privateKey,
+        coin: coinName,
+        network: networkType ?? NetworkEnum.Mainnet
+      })
+      return asPublicKey({
+        publicKeys: currencyFormatKeys
+      })
+    })()
+
```
swansontec, line 152: Given the problems with `asWalletKeysFormats`, change to:

```
const format = asOptionalCurrencyFormat(keys.format)
return {
  ...,
  keys: {
    format,
    supportedFormats: getSupportedFormats(format),
    ...
```

src/common/plugin/makeCurrencyTools.ts
```diff
@@ -30,36 +33,32 @@ export function makeCurrencyTools(
...
-        coin: engineInfo.network
-      })
       const mnemonic = bip39.entropyToMnemonic(Buffer.from(io.random(32)))
-      const keys: JsonObject = {
-        [mnemonicKey]: mnemonic
-      }
 
-      return {
-        ...keys,
+      const privateKey = asCurrencyPrivateKey({
```
swansontec, line 50: Use a type definition here, instead of a cleaner:

```
const privateKey: PrivateKey = {
  privateKey: mnemonic,
  format: opts?.format ?? engineInfo.formats?.[0] ?? 'bip44',
  coinType: opts?.coinType ?? engineInfo.coinType ?? 0
}
```

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -177,14 +189,15 @@ export async function makeUtxoEngine(
     },
 
     getDisplayPrivateSeed(): string | null {
-      return getMnemonic({ keys: walletInfo.keys, coin: engineInfo.network })
+      // Private key may be missing for watch-only wallets
+      const privateKey = asMaybe(asCurrencyPrivateKey)(sensitiveWalletInfo.keys)
+      if (privateKey == null)
```
swansontec, line 194: I think the intention was to return `null` here when unavailable, rather than throwing.

swansontec (changes requested):
A few more changes are needed, to avoid double-cleaning data.

samholmes:
Squashed the first two commits as suggested (rebasing the two commits into smaller changes was too time consuming). I added fixup commits to hopefully make it clearer that all the feedback has been taken into consideration.

src/common/utxobased/keymanager/cleaners.ts
```diff
@@ -0,0 +1,173 @@
...
+): Cleaner<NumbWalletInfo> => {
+  return (walletInfo: any) => {
+    const { engineInfo, coinInfo } = pluginInfo
+    const asCoinType = asOptional(asNumber, coinInfo.coinType)
+
+    // Prepare fields to return for NumbWalletInfo:
+    const { id, type } = walletInfo
+    const coinType = asCoinType(walletInfo.keys.coinType)
+    const format = asOptionalCurrencyFormat(walletInfo.keys.format)
+    const supportedFormats = getSupportedFormats(format)
```
swansontec, line 133: I didn't notice this before, but this logic won't work in watch-only mode. It assumes we have a private key, which isn't a given.

I think this function could be split into two sections like:

```js
const private = asMaybe(asCurrencyPrivateKey)(walletInfo.keys)
if (private != null) {
  // convert private to public
  return { format: private.format, ... }
}

const public = asMaybe(asCurrencyPublicKey)(walletInfo.keys)
if (public != null) {
  const formats = Object.keys(public.publicKeys).filter(asWalletFormat)
  return { format: formats[0], supportedFormats: formats, ... }
}

// If we get here, nothing matched:
throw new Error('either a public or private key are required')
```

This structure gets rid of the IIFE, and since we just try the keys both ways an go with whichever works.

This would allow `asNumbWalletInfo` to work in watch-only mode. Right now it will just blow up.

swansontec (changes requested):
More changes needed.

swansontec (approved):

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -158,12 +167,12 @@ export async function makeUtxoEngine(
 
     async dumpData(): Promise<EdgeDataDump> {
       return {
-        walletId: walletInfo.id.split(' - ')[0],
-        walletType: walletInfo.type,
+        walletId: sensitiveWalletInfo.id.split(' - ')[0],
+        walletType: sensitiveWalletInfo.type,
```
samholmes, line 171: Use the numbed type

src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -92,13 +91,15 @@ export function makeUtxoEngineState(
   const {
     network,
     pluginInfo,
-    walletInfo,
+    numbWalletInfo: walletInfo,
```
samholmes, line 94: Change it to just `walletInfo`

thehobbit85 (approved):

src/common/utxobased/keymanager/cleaners.ts
```diff
@@ -0,0 +1,189 @@
...
+      const formats = Object.entries(publicKey.publicKeys)
+        // Filter out undefined values in the entries because cleaners allow
+        // undefined values for optional fields.
+        .filter(([, value]) => value != null)
+        // Map to the entry's key as a currency format
+        .map(([format]) => asMaybe(asCurrencyFormat)(format))
+        // Filter out any formats that didn't pass the cleaner assertion
+        .filter(
+          (format?: CurrencyFormat): format is CurrencyFormat => format != null
+        )
```
samholmes, line 140: This had to be fixed. The formats needed to be changed from accessing the `Object.keys(publicKey.publicKeys)` to `Object.entries(publicKey.publicKeys)` so that way we can filter out the values (because they may be undefined).

PR #232, opened 2021-11-03
Sam/untangle coin info

samholmes:


samholmes:
We've decided not to do this because we want to make tests very simple: pass a string to the function instead of importing the coin info for the test.

PR #235, opened 2021-11-05
Keymanager: Support multiple prefixes from the same coin

TheCharlatan:
---
  - https://app.asana.com/0/0/1201238254464027

### DIFF
--- a/src/common/utxobased/info/dash.ts
+++ b/src/common/utxobased/info/dash.ts
@@ -61,12 +61,12 @@ export const coinInfo: CoinInfo = {
   segwit: false,
   coinType: 5,
   mainnetConstants: {
-    messagePrefix: 'unused',
-    wif: 0xcc,
-    legacyXPriv: 0x02fe52f8,
-    legacyXPub: 0x02fe52cc,
-    pubkeyHash: 0x4c,
-    scriptHash: 0x10
+    messagePrefix: ['unused'],
+    wif: [0xcc],
+    legacyXPriv: [0x02fe52f8],
+    legacyXPub: [0x02fe52cc],
+    pubkeyHash: [0x4c],
+    scriptHash: [0x10]
   }
 }
 
--- a/src/common/utxobased/info/ravencoin.ts
+++ b/src/common/utxobased/info/ravencoin.ts
@@ -52,13 +52,12 @@ export const coinInfo: CoinInfo = {
   segwit: false,
   coinType: 175,
   mainnetConstants: {
-    messagePrefix: '\x18Bitcoin Signed Message:\n',
-    wif: 0x80,
-    legacyXPriv: 0x0488ade4,
-    legacyXPub: 0x0488b21e,
-    pubkeyHash: 0x3c,
-    scriptHash: 0x7a,
-    bech32: 'bc'
+    messagePrefix: ['\x18Bitcoin Signed Message:\n'],
+    wif: [0x80],
+    legacyXPriv: [0x0488ade4],
+    legacyXPub: [0x0488b21e],
+    pubkeyHash: [0x3c],
+    scriptHash: [0x7a]
   }
 }
 
--- a/src/common/utxobased/keymanager/keymanager.ts
+++ b/src/common/utxobased/keymanager/keymanager.ts
@@ -5,6 +5,8 @@ import * as bip39 from 'bip39'
 import bitcoinMessage from 'bitcoinjs-message'
 import { InsufficientFundsError } from 'edge-core-js/types'
 
+import { indexAtProtected } from '../../../util/indexAtProtected'
+import { undefinedIfEmptyString } from '../../../util/undefinedIfEmptyString'
 import { CoinInfo, CoinPrefixes } from '../../plugin/types'
 import { IUTXO } from '../db/types'
 import { ScriptTemplate } from '../info/scriptTemplates/types'
@@ -94,7 +96,6 @@ export interface XPubToPubkeyArgs {
 export interface AddressToScriptPubkeyArgs {
   address: string
   addressType?: AddressTypeEnum
-  legacy?: boolean
   coin: string
 }
 
@@ -234,9 +235,9 @@ export interface Bip32 {
 
 interface Bip32NetworkFromCoinArgs {
   coinString: string
+  prefixIndex: number
   sigType?: BIP43PurposeTypeEnum
   forWIF?: boolean
-  legacy?: boolean
 }
 
 function bip43PurposeTypeEnumToNumber(purpose: BIP43PurposeTypeEnum): number {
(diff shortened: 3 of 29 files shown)


src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -471,14 +518,29 @@ export function seedOrMnemonicToXPriv(args: SeedOrMnemonicToXPrivArgs): string {
...
+    } catch (err) {
+      if (isError(err)) {
+        if (err.message === 'Invalid private key') {
+          prefixIndex += 1
+          continue
+        }
+      }
+      throw err
+    }
+  }
```
samholmes, line 543: Does this halt?

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -493,100 +555,133 @@ export function derivationLevelScriptHash(): number {
...
+    } catch (err) {
+      if (isError(err)) {
+        if (err.message === 'Invalid network version') {
+          prefixIndex += 1
+          continue
+        }
+      }
+      throw err
+    }
+  }
```
samholmes, line 583: Hmm, this loop pattern doesn't show a clear indication of halting. I'm wary of infinite loops cropping up if this code decays to that overtime. Perhaps a better approach would be to loop over all prefixes and throw outside that loop. This would make it clear that the iterations are bounded by the items in the set of derived prefixes by the `prefixIndex`.

samholmes (changes requested):
Need to discuss feedback

src/common/utxobased/keymanager/coins/bitcoin.ts
```diff
@@ -6,30 +6,30 @@ export class Bitcoin implements Coin {
...
+    wif: [0x80],
+    legacyXPriv: [0x0488ade4],
+    legacyXPub: [0x0488b21e],
+    wrappedSegwitXPriv: [0x049d7878],
+    wrappedSegwitXPub: [0x049d7cb2],
+    segwitXPriv: [0x04b2430c],
+    segwitXPub: [0x04b24746],
+    pubkeyHash: [0x00],
+    scriptHash: [0x05],
+    bech32: ['bc']
```
samholmes, line 19: These should include the legacy constants as the second element of each array. It's apart of the requirement for the task: 

> The datatype for each prefix field must be an array where the first element in the array is the default prefix for creating addresses.

samholmes (changes requested):
Another change needed to this work to meet the task requirements.

src/common/utxobased/info/dash.ts
```diff
@@ -60,12 +60,12 @@ export const coinInfo: CoinInfo = {
   segwit: false,
   coinType: 5,
   mainnetConstants: {
-    messagePrefix: 'unused',
-    wif: 0xcc,
-    legacyXPriv: 0x02fe52f8,
-    legacyXPub: 0x02fe52cc,
-    pubkeyHash: 0x4c,
-    scriptHash: 0x10
+    messagePrefix: ['unused'],
```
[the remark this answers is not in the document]
samholmes: If `messagePrefix` doesn't exist for a currency, hardcode to use a default that is bitcoin's messagePrefix.

samholmes (changes requested):
Partial review for now.

src/common/utxobased/info/ravencoin.ts
```diff
@@ -51,13 +51,13 @@ export const coinInfo: CoinInfo = {
...
-    pubkeyHash: 0x3c,
-    scriptHash: 0x7a,
-    bech32: 'bc'
+    messagePrefix: ['\x18Bitcoin Signed Message:\n'],
+    wif: [0x80],
+    legacyXPriv: [0x0488ade4],
+    legacyXPub: [0x0488b21e],
+    pubkeyHash: [0x3c],
+    scriptHash: [0x7a],
+    bech32: ['bc']
```
samholmes, line 60: Remove bech32. Ravencoin doesn't have segwit

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -269,64 +269,137 @@ export function bip43PurposeNumberToTypeEnum(
...
+  scriptHash: number
+  bech32?: string
+  cashaddr?: string
+}
+
+function selectedCoinPrefixes(
+  prefixes: CoinPrefixes,
+  index: number
+): SelectedCoinPrefixes {
+  const cleanedPrefixes = {
```
samholmes, line 291: Make cleaner/leaner.

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -483,149 +573,206 @@ export function derivationLevelScriptHash(
...
-          scriptType: ScriptTypeEnum.p2pkh,
-          coin: args.coin
-        })
+  const coin = getCoinFromString(args.coin)
+  const cashAddrPrefixes = getCashAddrPrefixes(coin)
+  const coinPrefixes = coin.mainnetConstants
+  const maxLength = getMaxLengthOfCoinPrefixes(coinPrefixes)
+  let scriptPubkey: string | undefined
+
+  for (let prefixIndex = 0; prefixIndex < maxLength; prefixIndex++) {
```
samholmes, line 615: Abstract the loop for all the function as a iterator function which accepts a callback.

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -780,17 +930,46 @@ export function pubkeyToScriptPubkey(
...
+  let max = 0
+  for (const value of Object.values(coin)) {
+    if (value instanceof Array) {
+      if (max < value.length) {
+        max = value.length
+      }
+    }
+  }
+  return max
+}
```
samholmes, line 943: Use `reduce`

src/common/utxobased/info/dash.ts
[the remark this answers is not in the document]
samholmes: Looks like this prefix is not option according to the `CoinPrefix` type. The reason for this is because AltcoinJS expects `messagePrefix` in its network type it would seem. Making it optional would only mean we need to add more logic to turn it back into an empty/bogus string, so we might as well leave it as a bogus string.

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -287,55 +289,22 @@ function selectedCoinPrefixes(
...
+    wrappedSegwitXPriv: indexAtProtected(prefixes.wrappedSegwitXPriv, index),
+    wrappedSegwitXPub: indexAtProtected(prefixes.wrappedSegwitXPub, index),
+    segwitXPriv: indexAtProtected(prefixes.segwitXPriv, index),
+    segwitXPub: indexAtProtected(prefixes.segwitXPub, index),
+    pubkeyHash: indexAtProtected(prefixes.pubkeyHash, index),
+    scriptHash: indexAtProtected(prefixes.scriptHash, index),
+    bech32: indexAtProtected(prefixes.bech32, index),
+    // Special handling for cashaddr to accommodate legacy addresses.
+    // By allowing them to overflow we will fallback to p2pkh.
+    cashaddr: undefinedIfEmptyString((prefixes.cashaddr ?? [])[index])
```
thehobbit85, line 306: Maybe I missed something, but I think this is basically the same:

```js
const selectedCoinPrefixes = (prefixes: CoinPrefixes, index: number): SelectedCoinPrefixes =>
  Object.entries(prefixes).reduce(
    (res, [type, prefixs = []]) => ({
      ...res,
      [type]: type !== 'cashaddr' 
        ? prefixs[Math.min(index, prefixs.length - 1)] 
        : prefixs[index] !== '' 
          ? prefixs[index] 
          : undefined
    }),
    {}
  )
```

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -899,15 +899,10 @@ export function pubkeyToScriptPubkey(
...
-  let max = 0
-  for (const value of Object.values(coin)) {
-    if (value instanceof Array) {
-      if (max < value.length) {
-        max = value.length
-      }
-    }
-  }
-  return max
+  return Object.values(coin).reduce(
```
thehobbit85, line 896: Since it's just a reduce, do we really even to make it a function?

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -387,33 +387,32 @@ function bip32NetworkFromCoin(
...
-      })
-      guessAddressTypeFromAddress(args.address, network, args.coin, undefined)
-      if (prefixIndex === 0) {
-        return VerifyAddressEnum.good
-      }
-      // All indices above 0 are treated as legacy
-      return VerifyAddressEnum.legacy
-    } catch (e) {
-      if (e instanceof Error) {
+  const coin = getCoinFromString(args.coin)
```
thehobbit85, line 390: This is not what I was thinking about, lets discuss it tomorrow.

thehobbit85 (changes requested):

src/common/utxobased/keymanager/keymanager.ts
samholmes: My original attempt was to use reduce. But I had concerns; which have been resolved with our discussion.

src/common/utxobased/keymanager/keymanager.ts
samholmes: I'll check and see. Good point.

src/common/utxobased/keymanager/keymanager.ts
samholmes: You're suggestion is awesome! Implement like this:

```
const xprivToXPubRaw = (args: XPrivToXPubArgs, prefixIndex: number): string => {
  const network: BitcoinJSNetwork = bip32NetworkFromCoin({
    coinString: args.coin,
    sigType: args.type,
    prefixIndex,
  })
  const coin = getCoinFromString(args.coin)
  const bip32FromBase58Func = coin.bip32FromBase58Func ?? bip32.fromBase58
  return bip32FromBase58Func(args.xpriv, network).neutered().toBase58()
}

export const xprivToXPub_ = (...arg):string => filterCoinPrefixes(xprivToXPubRaw, 'Invalid private key')(...arg)
```

src/common/utxobased/keymanager/keymanager.ts
samholmes: This fails test: 

```ts
function selectedCoinPrefixes(
  coinPrefixes: CoinPrefixes,
  index: number
): SelectedCoinPrefixes {
  const hack: SelectedCoinPrefixes = {} as any
  return Object.entries(coinPrefixes).reduce(
    (acc, [prefixType, prefixes]) => ({
      ...acc,
      [prefixType]:
        prefixType === 'cashaddr'
          ? undefinedIfEmptyString((prefixes.cashaddr ?? [])[index])
          : indexAtProtected(prefixes, index)
    }),
    hack
  )
}
```

Unfortunately, I spent too much time on this one so I'm pushing back on this change.

src/common/utxobased/keymanager/keymanager.ts
samholmes: Inlined it

thehobbit85 (approved):

PR #244, opened 2021-11-19
Dogecoin: Add min relay fee rate

TheCharlatan:
Dogecoin has a min relay fee of 0.001 Doge / KByte according to their newest fee scheme described in
https://github.com/dogecoin/dogecoin/discussions/2347 . Enforce this by adding a fee checker function to the CoiInfo.

---
  - https://app.asana.com/0/0/1201274042339100


TheCharlatan:
While the first commit adds simple logic, the second commit is more in tune with the asana task requirements of writing a `CoiInfo` "fee serializer". I committed both methods here to showcase the approaches.

samholmes (approved):
Squash commits before merging

PR #255, opened 2021-12-07
Sam/Fix Zero-Amount After Send

samholmes:
---
  - https://app.asana.com/0/0/1201447744301902


samholmes:
Feedback notes from previous review: Create and include the entire `IProcessorTransaction` type in `otherParams` on the `EdgeTransaction` returned by `makeSpend`. This way all that is necessary to convert to/from EdgeTransaction/IProcessorTransaction can be included in the `toEdgeTransaction` and `fromEdgeTransaction` functions.

thehobbit85 (approved):

PR #256, opened 2021-12-11
Sam/Fix Double Balance

samholmes:
---
  - https://app.asana.com/0/0/1201481239201205

### DIFF
--- a/src/common/utxobased/engine/makeUtxoEngineState.ts
+++ b/src/common/utxobased/engine/makeUtxoEngineState.ts
@@ -102,22 +102,22 @@ export function makeUtxoEngineState(
     addressWatching: false,
     blockWatching: false,
     addressSubscribeCache: {},
-    transactionsCache: {},
-    utxosCache: {},
-    rawUtxosCache: {},
-    processedUtxosCache: {},
-    updateTransactionsCache: {}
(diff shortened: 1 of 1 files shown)


src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -1144,6 +1144,11 @@ const processAddressUtxos = async (
       if (addressData == null || addressData.path == null) {
         return
       }
+
+      if (utxos.length === 0) {
+        addToProcessorUtxoCache(processorUtxoCache, path, scriptPubkey, 0)
```
samholmes, line 1149: Add `return` after

src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -792,12 +791,12 @@ export const pickNextTask = async (
   }
 
   // loop to get and process transaction history of single addresses, triggers setLookAhead
-  for (const address of Object.keys(transactionsCache)) {
-    const state = transactionsCache[address]
+  for (const address of Object.keys(addressTransactionCache)) {
```
samholmes, line 794: Sense we're switching to entries, might as well do it everywhere because we're touching these lines of code.

PR #258, opened 2021-12-16
Fix Sweep Private keys

samholmes:
---
  - https://app.asana.com/0/0/1201440626414086

### DIFF
--- a/src/common/utxobased/db/makeProcessor.ts
+++ b/src/common/utxobased/db/makeProcessor.ts
@@ -3,7 +3,7 @@ import * as bs from 'biggystring'
 import { Disklet, navigateDisklet } from 'disklet'
 import { EdgeGetTransactionsOptions } from 'edge-core-js/types'
 
-import { AddressPath } from '../../plugin/types'
+import { AddressPath, ChangePath } from '../../plugin/types'
 import { makeBaselets } from './makeBaselets'
 import { addressPathToPrefix, TxIdByDate } from './Models/baselet'
 import {
@@ -108,11 +108,9 @@ export interface Processor {
 
   saveAddress: (args: IAddress) => Promise<void>
   // used to calculate total number of addresses
-  numAddressesByFormatPath: (path: Omit<AddressPath, 'addressIndex'>) => number
+  numAddressesByFormatPath: (path: ChangePath) => number
   // get the last used address index for a specific format
-  lastUsedIndexByFormatPath: (
-    path: Omit<AddressPath, 'addressIndex'>
-  ) => Promise<number>
+  lastUsedIndexByFormatPath: (path: ChangePath) => Promise<number>
   fetchAddress: (args: AddressPath | string) => Promise<IAddress | undefined>
 
   /* Block processing
@@ -525,13 +523,11 @@ export async function makeProcessor(
       })
     },
 
-    numAddressesByFormatPath(path: Omit<AddressPath, 'addressIndex'>): number {
+    numAddressesByFormatPath(path: ChangePath): number {
       return baselets.all.scriptPubkeyByPath.length(addressPathToPrefix(path))
     },
 
-    async lastUsedIndexByFormatPath(
-      path: Omit<AddressPath, 'addressIndex'>
-    ): Promise<number> {
+    async lastUsedIndexByFormatPath(path: ChangePath): Promise<number> {
       const [addressIndex] = await baselets.address(async tables => {
         return await tables.lastUsedByFormatPath.query('', [
           addressPathToPrefix(path)
--- a/src/common/utxobased/db/makeProcessor.ts
+++ b/src/common/utxobased/db/makeProcessor.ts
@@ -260,7 +260,7 @@ export async function makeProcessor(
         // Return all UTXOs if no UTXO ids are specified
         if (utxoIds.length === 0) {
           const { data } = await tables.utxoById.dumpData('')
-          return Object.values(data[''])
+          return Object.values(data[''] ?? {})
         }
 
         return await tables.utxoById.query('', utxoIds)
--- a/src/common/utxobased/engine/makeUtxoEngine.ts
+++ b/src/common/utxobased/engine/makeUtxoEngine.ts
@@ -453,21 +453,35 @@ export async function makeUtxoEngine(
         coin: coinInfo.name
       })
 
-      const privateKeys = await Promise.all(
-        psbt.inputs.map(async ({ hash, index }) => {
-          const txid = Buffer.from(hash).reverse().toString('hex')
+      // Use the privateKeys (WIFs) from spendInfo, otherwise get them from the
+      // PBST inputs.
+      const privateKeys = await (async () => {
+        if (edgeSpendInfo.privateKeys != null) {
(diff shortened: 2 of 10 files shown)


src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -110,15 +110,21 @@ export function makeUtxoEngineState(
     updateTransactionCache: {}
   }
 
+  const clearObject = (obj: { [key: string]: any }): void => {
+    Object.keys(obj).forEach(key => {
+      // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+      delete obj[key]
+    })
+  }
```
thehobbit85, line 120: Why is this needed?
Doing something like this in Javascript is almost never needed and if it does, it usually means you are hiding some other issue.

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -492,28 +494,42 @@ export async function makeUtxoEngine(
...
+      ) => void
+      let failure: (err?: any) => void
+      const end = new Promise<EdgeTransaction>((resolve, reject) => {
+        success = resolve
+        failure = reject
+      })
+
+      // The temporary wallet info should include all formats for the currency
+      const allFormats = engineInfo.formats
+      if (allFormats == null) {
```
thehobbit85, line 520: Why do you need to check for null? I thought "engineInfo.formats" should be strictly typed to exists no?

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -523,53 +539,67 @@ export async function makeUtxoEngine(
...
-
-          const utxos = (await processor.fetchUtxos({ utxoIds: [] })) as IUTXO[]
-          if (utxos === null || utxos.length < 1) {
-            throw new Error('Private key has no funds')
+          try {
+            // Delay to ensure that UTXOs are processed.
+            // This is a bug in the engine state; addresses are considered
+            // processed before the UTXOs are saved.
+            // TODO: Remove delay when fixed.
+            await delay(1000)
```
thehobbit85, line 565: This should not even be as a temporary solution.
Lucky for us, the way to fix it is super simple, all you need to do is:

1. Add this `await args.onAddressChecked()` at line 1135 in 'makeUtxoEngineState.ts'
2. Change line 126 in  'makeUtxoEngineState.ts' to be: `processedCount += 0.5`

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -492,28 +494,42 @@ export async function makeUtxoEngine(
...
-      ) => void
-      let failure: (reason?: never) => void
-      const end: Promise<EdgeTransaction> = new Promise((resolve, reject) => {
-        success = resolve
-        failure = reject
-      })
-      const tmpDisklet = walletLocalDisklet
+
+      // Make temporary wallet disklet
+      const tmpDisklet = makeMemoryDisklet()
```
thehobbit85, line 542: No problem here. I actually wanted to say that this is very clever and I really like it :)

src/common/utxobased/db/makeProcessor.ts
```diff
@@ -260,6 +260,7 @@ export async function makeProcessor(
         // Return all UTXOs if no UTXO ids are specified
         if (utxoIds.length === 0) {
           const { data } = await tables.utxoById.dumpData('')
+          if (data[''] == null) return []
```
thehobbit85, line 263: You can do this instead: `return Object.values(data[''] ?? {})`

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -923,7 +923,7 @@ export async function makeTx(args: MakeTxArgs): Promise<MakeTxReturn> {
   const utxoPicker = coin.utxoPicker ?? utxopicker.utxoPicker
 
   const utxopicking =
-    args.forceUseUtxo != null ?? false
+    args.forceUseUtxo.length > 0
```
thehobbit85, line 926: You can always combine both tests into this: `(args.forceUseUtxo?.length ?? 0) > 0`

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -449,21 +449,32 @@ export async function makeUtxoEngine(
...
-
-          const address = await processor.fetchAddress(utxo.scriptPubkey)
-          if (address?.path == null) throw new Error('Invalid script pubkey')
-
-          return walletTools.getPrivateKey({ path: address.path, xprivKeys })
-        })
-      )
+      // Use the privateKeys (WIFs) from spendInfo, otherwise get them from the
+      // PBST inputs.
+      const privateKeys =
```
thehobbit85, line 454: This should not be a ternary. Just use the normal `let/if/else` pattern.

thehobbit85 (changes requested):

src/common/utxobased/engine/makeUtxoEngineState.ts
samholmes: This is to avoid reference errors. It's a fix to the poor implementation. Each cache object should be a Map object (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/clear). Though I as I understand, Maps introduce challenges on some targeted platforms.

src/common/utxobased/engine/makeUtxoEngine.ts
samholmes: No. The current type is optional.

src/common/utxobased/engine/makeUtxoEngine.ts
samholmes: Clever solution. Though I will want to add a pretty verbose comment explaining why processedCount is only incremented by 0.5.

src/common/utxobased/engine/makeUtxoEngine.ts
samholmes: Thanks! 👍

src/common/utxobased/db/makeProcessor.ts
samholmes: Yes. Nice suggestion!

src/common/utxobased/keymanager/keymanager.ts
samholmes: Beautiful!

src/common/utxobased/engine/makeUtxoEngine.ts
samholmes: I really wish we had `do` expressions. I dislike `let` for reasons. Can I use a IIFE with if/else?

thehobbit85: > IIFE

Definitely

src/common/utxobased/keymanager/keymanager.ts
samholmes: Wait, why do we need this? `forceUseUtxo` is never `undefined`.

thehobbit85 (approved):

PR #270, opened 2022-01-11
Fix BCH Blockbook Syncing

samholmes:
---
  - https://app.asana.com/0/0/1201590067175759


src/common/utxobased/engine/makeServerStates.ts
```diff
@@ -337,10 +337,6 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
           .then(response => {
             if (!resolved) {
               resolved = true
-              if ('error' in response) {
```
paullinator, line 340: This commit should have been broken up in two. One for the renames and one that actually changes code

thehobbit85 (approved):

PR #274, opened 2022-01-15
Sam/transaction spend amounts

samholmes:
~This is a hacky fix if we cannot figure out why Ravencoin is not sending new transactions for subscribed addresses~

Updated with a better solution to fixing this general problem.

---
  - https://app.asana.com/0/0/1201447744301902


paullinator:
At least make this for rvn only

samholmes:
@paullinator I considered that, but I have a better idea I'm going to try before moving any further with this solution.

thehobbit85 (approved):

PR #281, opened 2022-01-22
Fix local database configuration

samholmes:
---
  - https://app.asana.com/0/0/1201681616893842

### DIFF
--- a/src/common/utxobased/db/Models/baselet.ts
+++ b/src/common/utxobased/db/Models/baselet.ts
@@ -63,7 +63,7 @@ export interface TxIdByDate {
 export type TxIdByDateBaselet = RangeBase<TxIdByDate, 'date', 'txid'>
 export const txIdsByDateOptions: RangeBaseOptions<'date', 'txid'> = {
   name: 'txIdsByDate',
-  bucketSize: 30 * 24 * 60 * 60 * 1000,
+  bucketSize: 30 * 24 * 60 * 60, // 30 days
   rangeKey: 'date',
   idKey: 'txid'
 }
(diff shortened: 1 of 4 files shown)


src/common/utxobased/db/Models/baselet.ts
```diff
@@ -22,7 +22,7 @@ export const scriptPubkeyByPathOptions: CountBaseOptions = {
 export type AddressByScriptPubkeyBaselet = HashBase<IAddress>
 export const addressByScriptPubkeyOptions: HashBaseOptions = {
   name: 'addressByScriptPubkey',
-  prefixSize: 2
+  prefixSize: 6
```
paullinator, line 25: Shouldn't this be 4?

src/common/utxobased/db/Models/baselet.ts
```diff
@@ -74,14 +74,8 @@ export const utxoByIdOptions: HashBaseOptions = {
...
-export const blockHashByBlockHeightOptions: HashBaseOptions = {
-  name: 'blockHashByBlockHeight',
-  prefixSize: 2
-}
-
 export type UtxoIdsByScriptPubkeyBaselet = HashBase<string[]>
 export const utxoIdsByScriptPubkeyOptions: HashBaseOptions = {
   name: 'utxoIdsByScriptPubkey',
-  prefixSize: 2
+  prefixSize: 6
```
paullinator, line 80: Also 4

src/common/utxobased/db/Models/baselet.ts
paullinator: we don't ever use p2pk so 6 digit prefix is unnecessary

paullinator (changes requested):

paullinator (approved):

PR #282, opened 2022-01-22
Implement `importPrivateKey`

samholmes:
---
  - https://app.asana.com/0/0/1201689776652872


thehobbit85 (changes requested):
There is a problem with this implementation.
This "seedOrMnemonic" should be valid:
`d2149cdee8a731a953ee3d41fb6178745ec4d330454dde96bcb793b534dc2145`

samholmes:
Change to not adding a default value for `imported`. This way we always have the type `undefined | boolean` internallyl.

thehobbit85 (approved):

PR #292, opened 2022-02-23
Fix RBF

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

- https://github.com/EdgeApp/edge-react-gui/pull/4547

### Description

 none

---
  - https://app.asana.com/0/0/1201689712867366


src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -132,6 +140,209 @@ export async function makeUtxoEngine(
   })
 
   const engine: EdgeCurrencyEngine = {
+    async accelerate(edgeTx: EdgeTransaction): Promise<EdgeTransaction | null> {
+      // Get the replaced transaction rom the processor:
```
swansontec, line 144: s/rom/from/

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -132,6 +140,209 @@ export async function makeUtxoEngine(
...
+      })
+
+      // Transaction checks:
+      // The transaction must be found and not confirmed or dropped.
+      if (replacedTx == null) return null
+      if (replacedTx.blockHeight !== 0) return null
+
+      // Double the fee used for the RBF transaction:
+      const vBytes = transactionSizeFromHex(replacedTx.hex)
+      const newFeeRate = Math.round(parseInt(replacedTx.fees) / vBytes) * 2
```
swansontec, line 157: Put the `* 2` inside the `Math.round`, so we do the doubling in float-land instead of integer-land.

Rounding creates ±0.5 of "random noise" in the output values, but the `* 2` doubles this noise to ±1. If you perform the `* 2` before rounding, though, you only get the ±0.5 of noise needed to hit an integer, and nothing more.

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -132,6 +140,209 @@ export async function makeUtxoEngine(
...
+          targets.push({
+            scriptPubkey: output.scriptPubkey,
+            value: parseInt(output.amount)
+          })
+          continue
+        }
+
+        // The output is ours:
+        // Addresses from database should include a path; this is a type assert
+        if (ourAddress.path != null) {
```
swansontec, line 197: If this fails, the output won't be in `newOurReceiveAddresses`, `foundChangeAddress`, or `targets`. Is this going to be a problem? Would it be safer to treat the address as an external target if this check fails?

swansontec (changes requested):
I found some small things. This looks pretty good overall, with a lot of little changes that make sense individually.

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -73,6 +74,17 @@ export enum ScriptTypeEnum {
   replayProtection = 'replayprotection',
   replayProtectionP2SH = 'replayprotectionp2sh'
 }
+export const asScriptTypeEnum = asValue(
```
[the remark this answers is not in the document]
samholmes: TypeScript is dumb and wont work with this when it makes perfect sense that it should work.

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -428,7 +427,7 @@ export async function makeUtxoEngine(
         feeRate,
         coin: coinInfo.name,
         currencyCode: currencyInfo.currencyCode,
-        setRBF,
+        enableRbf: edgeSpendInfo.otherParams?.enableRbf ?? false,
```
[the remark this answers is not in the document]
samholmes: Break-away task for this one.

src/common/utxobased/keymanager/types.ts
```diff
@@ -0,0 +1,33 @@
...
+  networkFeeShortage?: string
+}
+export class InsufficientFundsErrorPlus extends InsufficientFundsError {
+  networkFeeShortage?: string
+  constructor(opts: InsufficientFundsErrorOptsPlus = {}) {
+    super(opts)
+    const { networkFeeShortage } = opts
+    this.networkFeeShortage = networkFeeShortage
+  }
+}
```
swansontec, line 18: Nice.

swansontec (approved):

PR #310, opened 2022-05-31
Make blockbook server URL consistent

paullinator:
Standardize on URL not including `/websocket` and defining it's own protocol. ie `wss:` or `ws:`

This allows us to test against unencrypted localhost


src/common/plugin/pluginState.ts
```diff
@@ -16,8 +16,8 @@ const SERVER_CACHE_FILE = 'serverCache.json'
 
 // Normalizes a URL to always be a websocket URL
 const asWebsocketUrl = (raw: unknown): string => {
-  const url = new URL(asString(raw))
-  return `wss://${url.host}/websocket`
+  const url = asString(raw)
+  return `${url}/websocket`
```
samholmes, line 20: Because of this change, all the info files need to be audit and the protocols need to be changed. The previous implementation "cleaned" every URL into the same shape, so it wouldn't matter if there was a `/websocket` suffix or not. Sense we're being stricter on the type, we need to regression test on all the currencies to make sure they sync.

samholmes (changes requested):
Just make sure the info files are correct after this change.

src/common/plugin/pluginState.ts
paullinator: I did audit them and you'll notice I changed the ones that were wrong.

samholmes (approved):

PR #311, opened 2022-06-02
Update the `userSettings` format

swansontec:
We don't want to read old electrum settings from the old plugin, since those won't work. We also want to change `disableFetchingServers` into the less-confusing `enableCustomServers`.

### DIFF
--- a/src/common/plugin/makeCurrencyPlugin.ts
+++ b/src/common/plugin/makeCurrencyPlugin.ts
@@ -8,6 +8,7 @@ import {
 } from 'edge-core-js/types'
 
 import { makeUtxoEngine } from '../utxobased/engine/makeUtxoEngine'
+import { asUtxoUserSettings } from '../utxobased/engine/types'
 import { makeCurrencyTools } from './makeCurrencyTools'
 import { makeEngineEmitter } from './makeEngineEmitter'
 import { makePluginState } from './pluginState'
@@ -21,22 +22,13 @@ export function makeCurrencyPlugin(
   const { io, log, pluginDisklet } = pluginOptions
   const currencyTools = makeCurrencyTools(io, pluginInfo)
   const { defaultSettings, pluginId, currencyCode } = currencyInfo
-  const {
-    customFeeSettings,
-    serverList,
-    disableFetchingServers
-  } = defaultSettings
   const state = makePluginState({
     io,
     currencyCode,
     pluginId,
     pluginDisklet,
     log,
-    defaultSettings: {
-      customFeeSettings,
-      serverList,
-      disableFetchingServers
-    }
+    defaultSettings: asUtxoUserSettings(defaultSettings)
   })
   return {
     currencyInfo,
--- a/src/common/plugin/pluginState.ts
+++ b/src/common/plugin/pluginState.ts
@@ -6,6 +6,7 @@ import { EdgeIo, EdgeLog } from 'edge-core-js/types'
 import { makeMemlet } from 'memlet'
 
 import { UtxoEngineState } from '../utxobased/engine/makeUtxoEngineState'
+import { UtxoUserSettings } from '../utxobased/engine/types'
 import { asServerInfoCache, ServerScores } from './serverScores'
 
 // Info server endpoint to getting ServerListInfo data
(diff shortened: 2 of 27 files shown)


src/common/utxobased/engine/types.ts
```diff
@@ -1,7 +1,20 @@
+import { asArray, asBoolean, asObject, asOptional, asString } from 'cleaners'
 import { EdgeSpendInfo } from 'edge-core-js/types'
 
 import { Input, Output } from '../keymanager/utxopicker/types'
 
+// Normalizes a URL to always be a websocket URL
+const asWebsocketUrl = (raw: unknown): string => {
```
swansontec, line 7: One concern I have about doing this here is that it means the engine could become un-bootable if the user provides invalid user settings. We may want to either:

1. Go back to a plain `asString` and just let the connection fail if it's not a valid URL. Server score would go to 0, and all would be well.
2. Filter the invalid entries out, either here in the cleaner or later when this is used.

I prefer option 1, since the normalization that's happening here is over-agressive and kinda daft.

src/common/utxobased/info/badcoin.ts
```diff
@@ -16,8 +15,8 @@ const currencyInfo: EdgeCurrencyInfo = {
   // Configuration options:
   defaultSettings: {
     customFeeSettings: ['satPerByte'],
-    ...serverListFields([]),
-    disableFetchingServers: true
+    blockbookServers: [],
+    enableCustomServers: false
```
swansontec, line 19: Notice that the default was wrong in the original code!

src/common/plugin/pluginState.ts
```diff
@@ -142,7 +135,8 @@ export function makePluginState(settings: PluginStateSettings): PluginState {
   const refreshServers = async (): Promise<void> => {
     let serverList = defaultServers
 
-    if (!disableFetchingServers)
+    // Does this logic seem backwards?
+    if (enableCustomServers)
```
swansontec, line 139: Oops, looking at the diff, I see I deleted the `!`, which is indeed why the logic is backwards now. Need to fix that.

src/common/utxobased/engine/types.ts
```diff
@@ -1,7 +1,20 @@
...
+  const url = new URL(asString(raw))
+  return `wss://${url.host}/websocket`
+}
+
+export const asUtxoUserSettings = asObject({
+  blockbookServers: asOptional(asArray(asWebsocketUrl), []),
+  enableCustomServers: asOptional(asBoolean, false)
+})
+export type UtxoUserSettings = ReturnType<typeof asUtxoUserSettings>
```
swansontec, line 16: Mirroring the concern above, should these be `asMaybe` instead of `asOptional`, so it's fool-proof?

swansontec (commented):
Here are some worries I have about my code.

src/common/plugin/makeCurrencyPlugin.ts
```diff
@@ -21,22 +22,13 @@ export function makeCurrencyPlugin(
...
     currencyCode,
     pluginId,
     pluginDisklet,
     log,
-    defaultSettings: {
-      customFeeSettings,
-      serverList,
-      disableFetchingServers
-    }
+    defaultSettings: asUtxoUserSettings(defaultSettings)
```
samholmes, line 31: Why are we cleaning the `defaultSettings` here if it's coming from `currencyInfo` which should have a trusted type?

src/common/utxobased/engine/types.ts
```diff
@@ -1,7 +1,20 @@
+import { asArray, asBoolean, asObject, asOptional, asString } from 'cleaners'
 import { EdgeSpendInfo } from 'edge-core-js/types'
 
 import { Input, Output } from '../keymanager/utxopicker/types'
 
+// Normalizes a URL to always be a websocket URL
+const asWebsocketUrl = (raw: unknown): string => {
+  const url = new URL(asString(raw))
+  return `wss://${url.host}/websocket`
+}
```
samholmes, line 10: We should de-dup this code in the `pluginState.js` file

samholmes (changes requested):

src/common/utxobased/engine/types.ts
samholmes: Instead: remove in favor of adding the `/websocket` where it's needed in the call-stack.

PR #313, opened 2022-06-04
Fix websocket URLs by bringing back cleaner asWebsocketUrl cleaner

samholmes:


src/common/utxobased/engine/types.ts
```diff
@@ -3,8 +3,14 @@ import { EdgeSpendInfo } from 'edge-core-js/types'
 
 import { Input, Output } from '../keymanager/utxopicker/types'
 
+// Normalizes a URL to always be a websocket URL
+export const asWebsocketUrl = (raw: unknown): string => {
+  const url = asString(raw)
+  return `${url}/websocket`
```
swansontec, line 9: You could use the safer regexp approach to only add this if needed, as isn’t the deleted code.

swansontec (approved):
Approved, but we could keep the regexp, just inside the cleaner.

paullinator:
dont merge

paullinator (changes requested):
Please change implementation to not add `websocket` to the URL. Instead only add it when you're opening the socket. Otherwise this will get included in the server name when stored on disk

PR #317, opened 2022-06-09
Sam/confirmations

samholmes:


src/common/utxobased/db/Models/ProcessorTransaction.ts
```diff
@@ -95,9 +97,21 @@ export const toEdgeTransaction = async (
     }
   }
 
+  const confirmations: EdgeTransaction['confirmations'] =
```
thehobbit85, line 101: Using the tx.blockHeight is incorrect. It should be similar to the way it's done in the core:

Where you use the `currentWalletBlockHeight - tx.blockHeight`

thehobbit85 (changes requested):

thehobbit85 (changes requested):
Also, remove the last commit

PR #323, opened 2022-06-17
Sam/fixes

samholmes:


src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -736,55 +731,42 @@ export const pickNextTask = async (
...
+      if (!serverStates.serverIsAwareOfAddress(uri, address)) {
+        newAddress.push(address)
       }
+
+      // Add to the addressTransactionCache if they're not yet added:
+      // Only process newly watched addresses
+      if (state.processing) continue
+      // Add the newly watched addresses to the UTXO cache
+      addressUtxoCache[address] = {
+        processing: false,
```
thehobbit85, line 765: Why do we even need this processing now?
Is it used somewhere else in the code that's not part of the commit?

Since it was used before on line 747:

```js 
const { path, processing: subscribed } = state
// only process newly watched addresses
if (subscribed) continue
```

but now it's not used (at least in the code here) anymore.

src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -736,55 +731,42 @@ export const pickNextTask = async (
...
-    serverStates.watchAddresses(
-      uri,
-      Array.from(Object.keys(addressSubscribeCache)),
-      deferred
-    )
-    return true
+    // Subscribe to any new addresses
+    if (newAddress.length > 0) {
+      serverStates.watchAddresses(uri, newAddress)
+      return true
```
thehobbit85, line 782: Is there a different meaning if the pickNextTask returns an undefined or true?
If so, you might want to move the true after the if.

But again, I don't fully understand how the return value from pickNextTask is used.

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -120,7 +120,10 @@ export async function makeUtxoEngine(
         metadata.balance
       )
 
+      // Refresh the servers for the engine
       pluginState.addEngine(state)
+      await pluginState.refreshServers()
```
thehobbit85, line 125: This should not be here.
It should also not be in the resync method.

Well, kinda, it should be inside the pluginState.clearCache method like here: https://github.com/EdgeApp/edge-currency-bitcoin/blob/61b56f73e871f2fffccc55e800ad7cfabf24897d/src/plugin/pluginState.js#L140

Moving it here should not be the solution to the race condition you're having. 
I would like to go over this with you tomorrow to try and find a different solution.

src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -111,12 +111,24 @@ export function makeUtxoEngineState(
...
+    }
+    for (const key of Object.keys(taskCache.rawUtxoCache)) {
+      removeItem(taskCache.rawUtxoCache, key)
+    }
+    for (const key of Object.keys(taskCache.processorUtxoCache)) {
+      removeItem(taskCache.processorUtxoCache, key)
+    }
+    for (const key of Object.keys(taskCache.updateTransactionCache)) {
+      removeItem(taskCache.updateTransactionCache, key)
+    }
```
thehobbit85, line 131: Since you are removing everything anyway, there is no benefit in doing this other then maybe type safety but it's not really needed here anyway.
Revert this commit please.

thehobbit85 (changes requested):

thehobbit85:
Also, cleanup the commits, you have a revert of a revert ;)

src/common/utxobased/engine/makeUtxoEngineState.ts
samholmes: the processing flag on the `addressSubscribedCache` entry is to signal whether we have added an entry to `addressUtxoCache` already.

src/common/utxobased/engine/makeUtxoEngineState.ts
samholmes: `true` means to continue to pick next task immediately after finishing; `false` or `undefined` exits from the `doWakeUp` loop (I'm assuming this is a sort of going to sleep mechanism).

The reason why we need to return in the `if` is because the outer block is not a condition which will ever be false, so it will halt the entire pick next task loop.

src/common/utxobased/engine/makeUtxoEngine.ts
samholmes: Let's discuss this point.

src/common/utxobased/engine/makeUtxoEngineState.ts
samholmes: It's needed to eliminate object reference bugs.

thehobbit85 (approved):

PR #336, opened 2022-07-08
Fail gracefully when fetching for blockbook server list

samholmes:
A fail here will cause the initial load to break especially in airplane
mode.


samholmes:
OTS approved

PR #338, opened 2022-07-20
Remove await for updateServers in makeUtxoServer unblocking wallet load

samholmes:


samholmes:
Approved by @paullinator while fixing OTS.

PR #339, opened 2022-07-20
Add OP_RETURN to outputs

GoodluckH:
The added logics check if the `target` has memo, if so, we know it's 
an `OP_RETURN` script. Then, we use the altcoin.js lib to construct
an `OP_RETURN` script value.

However, if the memo (in hex string) exceeds the 80 bytes limitation
for `OP_RETURN` script, we throw.

---
  - https://app.asana.com/0/0/1201934696774649

### DIFF
--- a/src/common/utxobased/engine/makeUtxoEngine.ts
+++ b/src/common/utxobased/engine/makeUtxoEngine.ts
@@ -311,7 +311,8 @@ export async function makeUtxoEngine(
             )
             targets.push({
               address: publicAddress,
-              value: parseInt(target.nativeAmount)
+              value: parseInt(target.nativeAmount),
+              memo: target.memo
             })
           }
         } else {
@@ -323,7 +324,8 @@ export async function makeUtxoEngine(
           }
           targets.push({
             address: target.publicAddress,
-            value: parseInt(target.nativeAmount)
+            value: parseInt(target.nativeAmount),
+            memo: target.memo
           })
         }
       }
--- a/src/common/utxobased/engine/makeUtxoEngine.ts
+++ b/src/common/utxobased/engine/makeUtxoEngine.ts
@@ -376,7 +376,7 @@ export async function makeUtxoEngine(
       log.warn(`spend: Using fee rate ${feeRate} sat/B`)
       const subtractFee =
         options?.subtractFee != null ? options.subtractFee : false
-      const tx = await makeTx({
+      const tx = makeTx({
         utxos,
         forceUseUtxo: maxUtxo != null ? [maxUtxo] : [],
         targets,
--- a/src/common/utxobased/engine/makeUtxoEngine.ts
+++ b/src/common/utxobased/engine/makeUtxoEngine.ts
@@ -297,7 +297,10 @@ export async function makeUtxoEngine(
       let targets: MakeTxTarget[] = []
       const ourReceiveAddresses: string[] = []
       for (const target of spendTargets) {
-        if (target.publicAddress == null || target.nativeAmount == null) {
+        if (
+          target.memo == null &&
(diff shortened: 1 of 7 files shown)


src/common/utxobased/engine/utils.ts
```diff
@@ -35,6 +35,13 @@ export const getCurrencyFormatFromPurposeType = (
   }
 }
 
+export const getNumberOfBytesFromHex = (hex: string): number => {
```
thehobbit85, line 38: This is very wasteful.
If you're string is already a valid hex sting, then the num of bytes is just `Math.ceil(hex.length / 2)`

src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -311,7 +311,8 @@ export async function makeUtxoEngine(
             )
             targets.push({
               address: publicAddress,
-              value: parseInt(target.nativeAmount)
+              value: parseInt(target.nativeAmount),
+              memo: target.memo == null ? undefined : target.memo
```
thehobbit85, line 315: No need to check if it's null here, just add it directly.
You are testing for null anyway in the createTx function anyway.
About the type issue, see my next comment.

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -191,6 +192,7 @@ export interface MakeTxArgs {
 export interface MakeTxTarget {
   address: string
   value: number
+  memo?: string
```
thehobbit85, line 195: Change this type to `memo?: string | null` to fix the type issues you have

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -904,10 +906,20 @@ export async function makeTx(args: MakeTxArgs): Promise<MakeTxReturn> {
...
+    let script = addressToScriptPubkey({
       address: target.address,
       coin: coin.name
     })
+    if (target.memo != null) {
+      const memoHex = Buffer.from(target.memo, 'utf8').toString('hex')
+      // check if hex string is within 80 bytes
+      if (getNumberOfBytesFromHex(memoHex) > 80) {
+        throw new Error('Memo size exceeds 80 bytes')
+      }
```
thehobbit85, line 918: This should be written as the 'validateMemo' function in the makeCurrencyTools following this typing - https://github.com/EdgeApp/edge-core-js/blob/afbdb5ece2964ea09c29fb28b816ad8d9e726b19/src/types/types.js#L670

thehobbit85 (changes requested):

src/common/utxobased/keymanager/keymanager.ts
thehobbit85: This is still not resolved, you should put the function on the currencyTools object.

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -191,6 +192,7 @@ export interface MakeTxArgs {
 export interface MakeTxTarget {
   address: string
   value: number
+  memo?: string | null
```
GoodluckH, line 195: Address and value should also be optional.

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -903,16 +905,29 @@ export async function makeTx(args: MakeTxArgs): Promise<MakeTxReturn> {
...
+        .toString('hex')
+    }
+    targets.push({
       script,
       value: target.value
-    }
-  })
+    })
+  }
+
```
GoodluckH, line 930: OpReturn should be an additional target with value of zero

thehobbit85 (approved):

PR #340, opened 2022-08-17
WIP Always delete a UTXO that's missing from the network

paullinator:
---
  - https://app.asana.com/0/0/1202817248343315


samholmes (commented):
I say we get this on a cheese built to have heavily tested by QA, or we do some edge-case dev testing simulating disparaged networks in order to see if we can produce a double-balance bug. If we're unable to produce the double-balance bug, then we can say this solution is safe to merge as an intermediary fix.

PR #343, opened 2022-08-23
Add outputSort param to makeSpend

paullinator:
Allows sorting by bip69 convention or in order of targets passed in with change as last output.

Unit tests verified by decoding signedTx using web decoder to ensure outputs are properly sorted given the outputSort param value.

---
  - https://app.asana.com/0/0/1202844141125919

### DIFF
--- a/test/common/utxobased/keymanager/coins/bitcointransactiontest.spec.ts
+++ b/test/common/utxobased/keymanager/coins/bitcointransactiontest.spec.ts
@@ -186,7 +186,8 @@ describe('bitcoin transaction creation and signing test', function () {
           vout: 0
         }
       ],
-      targets: []
+      targets: [],
+      outputSort: 'bip69'
     })
     const signedTx = await signTx({
       psbtBase64,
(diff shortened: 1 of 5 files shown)


test/common/utxobased/keymanager/coins/bitcointransactiontest.spec.ts
```diff
@@ -203,6 +204,128 @@ describe('bitcoin transaction creation and signing test', function () {
...
+      ],
+      outputSort: 'targets'
+    })
+    const signedTx = await signTx({
+      psbtBase64,
+      privateKeys: [privateKey],
+      coin: 'bitcoin'
+    })
+    expect(signedTx.hex).to.equal(
+      '02000000013ebc8203037dda39d482bf41ff3be955996c50d9d4f7cfc3d2097a694a7b067d000000006a47304402203d4b4d8f7be9219914e822684ce72001367081101654aaeeaa734e6aa934972e022072154039d101819e8c78fc3a11132e5a46a3538546db68c8a7818128b02de88101210365db9da3f8a260078a7e8f8b708a1161468fb2323ffda5ec16b261ec1056f455ffffffff042c010000000000001600148bbc95d2709c71607c60ee3f097c1217482f518dc8000000000000001600148bbc95d2709c71607c60ee3f097c1217482f518d64000000000000001600148bbc95d2709c71607c60ee3f097c1217482f518d28360100000000001976a914ca0d36044e0dc08a22724efa6f6a07b0ec4c79aa88ac00000000'
```
samholmes, line 325: We should really cleanup these tests to expect some more readable output fixture data. How are we to know this output hex data is correct? I only request writing up a breakaway task to fix this.

samholmes (approved):

test/common/utxobased/keymanager/coins/bitcointransactiontest.spec.ts
paullinator: I actually decoded the output value using https://live.blockcypher.com/btc/decodetx/ and verified the order of the outputs

paullinator:
Holding off merge until 2.20.2 is released.

PR #346, opened 2022-10-01
Add getTransactionSpecific endpoint support

peachbits:
The "transaction specific" blockbook functionality include additional data that is unique to the network the transaction is on. For example, this is where Dash's InstantSend boolean is returned since that wouldn't be appropriate in the general transaction endpoint. 

This PR adds the following:
1) Add transaction specific functionality
2) Use new endpoint and the core's confirmation API to update pending InstantSend transactions as confirmed
3) Add logic to avoid the extra query and engine callback if the plugin doesn't require the new method

---
  - https://app.asana.com/0/0/1201425242712367

### DIFF
--- a/src/common/plugin/types.ts
+++ b/src/common/plugin/types.ts
@@ -10,7 +10,7 @@ import {
   EdgeWalletInfo
 } from 'edge-core-js/types'
 
-import { IUTXO } from '../utxobased/db/types'
+import { IProcessorTransaction, IUTXO } from '../utxobased/db/types'
 import { ScriptTemplates } from '../utxobased/info/scriptTemplates/types'
 import { UtxoPicker } from '../utxobased/keymanager/utxopicker'
 import { EngineEmitter } from './makeEngineEmitter'
@@ -56,6 +56,11 @@ export interface EngineInfo {
   scriptTemplates?: ScriptTemplates
   // Codec Cleaners
   asBlockbookAddress?: Cleaner<string>
+  // Coin specific transaction handling
+  txSpecificHandling?: (
+    tx: IProcessorTransaction,
+    specialTx: unknown
+  ) => IProcessorTransaction
 }
 
 /**
--- a/src/common/utxobased/engine/makeUtxoEngineState.ts
+++ b/src/common/utxobased/engine/makeUtxoEngineState.ts
@@ -678,7 +678,7 @@ export const pickNextTask = async (
   args: NextTaskArgs
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
 ): Promise<WsTask<any> | undefined | boolean> => {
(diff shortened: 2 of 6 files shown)


src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -977,9 +1004,11 @@ const updateTransactions = (
         tx: processedTx
       })
 
-      // Add task to grab transactionSpecific payload
-      taskCache.updateTransactionSpecificCache[txId] = {
-        processing: false
+      if (needsTxSpecific) {
```
samholmes, line 1007: Should we inline the condition here rather than thread through a parameter?

src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -856,6 +898,47 @@ interface UpdateTransactionsArgs extends CommonArgs {
   txId: string
 }
 
+const updateTransactionsSpecific = (
+  args: UpdateTransactionsArgs
```
samholmes, line 916: It should have it's own args type.

samholmes (changes requested):

src/common/utxobased/engine/makeUtxoEngineState.ts
samholmes: Nah, we'll keep it wired through. There's an ongoing repayment of tech-debt to make these arg-bag parameters saner.

src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -691,6 +691,14 @@ export const pickNextTask = async (
     updateTransactionSpecificCache
   } = taskCache
 
+  /**
+   * Some currencies require an additional blockbook payload 'getTransactionSpecific' in order
+   * to provide all relevant transaction data. Since this is currency-specific, we can limit
+   * the useage to currencies that require it.
+   **/
+
```
samholmes, line 699: Nit: Remove newline

samholmes (changes requested):

src/common/plugin/types.ts
```diff
@@ -56,6 +57,11 @@ export interface EngineInfo {
   scriptTemplates?: ScriptTemplates
   // Codec Cleaners
   asBlockbookAddress?: Cleaner<string>
+  // Coin specific transaction handling
+  txSpecificHandling?: (
```
samholmes, line 61: Alternatively, you can make the `specialTx: TransactionSpecificResponse` an `unknown` field.

samholmes (changes requested):

samholmes (approved):
Please make sure the commits look clean after squash.

PR #347, opened 2022-10-26
New wallet balance calculation algorithm

samholmes:
---
  - https://app.asana.com/0/0/1203093844594819


swansontec (approved):
I had some questions at first, but then the later commits answered them, so this looks good.

PR #349, opened 2022-11-20
Features needed for Thorchain Savers

paullinator:
This is blocked on PR https://github.com/EdgeApp/edge-core-js/pull/489 which needs to be merged and published and this PR will need an upgrade of the core before merge

---
  - https://app.asana.com/0/0/1203395121961365

### DIFF
--- a/src/common/utxobased/engine/makeUtxoEngine.ts
+++ b/src/common/utxobased/engine/makeUtxoEngine.ts
@@ -284,7 +284,13 @@ export async function makeUtxoEngine(
       options?: TxOptions
     ): Promise<EdgeTransaction> {
       const { spendTargets } = edgeSpendInfo
-      const { outputSort = 'bip69' } = edgeSpendInfo.otherParams ?? {}
+      const { outputSort = 'bip69', utxoSourceAddress, forceChangeAddress } =
+        edgeSpendInfo.otherParams ?? {}
+
+      let utxoScriptPubkey: string | undefined
+      if (utxoSourceAddress != null) {
+        utxoScriptPubkey = walletTools.addressToScriptPubkey(utxoSourceAddress)
+      }
 
       if (options?.CPFP == null && spendTargets.length < 1) {
         throw new Error('Need to provide Spend Targets')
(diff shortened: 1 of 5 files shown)


src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -296,7 +302,10 @@ export async function makeUtxoEngine(
       )
       const utxos =
         options?.utxos ??
-        ((await processor.fetchUtxos({ utxoIds: [] })) as IUTXO[])
+        ((await processor.fetchUtxos({
+          scriptPubkey: utxoScriptPubkey,
+          utxoIds: []
+        })) as IUTXO[])
```
samholmes, line 308: The type assertion isn't trustworthy anymore now that we've parameterized `fetchUtxo` differently. Use `filterUndefined` utility function to actually remove undefined values from the return value.

src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -352,6 +352,7 @@ export function makeUtxoEngineState(
         })
 
         return {
+          nativeAmount: '0',
```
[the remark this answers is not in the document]
samholmes: Oh I see, this is for unreleased core changes (hence the draft status for the PR). Never mind these comments.

samholmes (changes requested):
All of these changes make sense to me. I only have one change request from my first run-through with this PR.

src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -352,16 +356,27 @@ export function makeUtxoEngineState(
...
+        } = await internalGetFreshAddress({
           ...commonArgs,
           format: getCurrencyFormatFromPurposeType(BIP43PurposeTypeEnum.Segwit),
           forceIndex,
           changeIndex: branch
         })
 
         return {
+          nativeBalance,
+          segwitNativeBalance,
```
samholmes, line 371: Sort these where the balance follows address for each type. Also this commit's message looks like it needs some minor cleanup.

samholmes (changes requested):
nativeBalance is a good name given that nativeAmount is reserved for something else. I added additional minor feedback to clean up some things and the other review's feedback item remains.

samholmes (approved):
Approved and I squashed the commits.

PR #356, opened 2022-12-29
Support bech32 addresses

samholmes:
### Dependencies

- https://github.com/EdgeApp/edge-core-js/pull/494

---
  - https://app.asana.com/0/0/1203136033933205

### DIFF
--- a/src/common/utxobased/engine/makeUtxoEngineState.ts
+++ b/src/common/utxobased/engine/makeUtxoEngineState.ts
@@ -26,7 +26,7 @@ import {
   IUTXO,
   makeIAddress
 } from '../db/types'
-import { NumbWalletInfo } from '../keymanager/cleaners'
+import { getSupportedFormats, NumbWalletInfo } from '../keymanager/cleaners'
 import {
   BIP43PurposeTypeEnum,
   derivationLevelScriptHash,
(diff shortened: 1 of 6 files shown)


src/common/utxobased/engine/makeUtxoEngineState.ts
```diff
@@ -345,73 +343,54 @@ export function makeUtxoEngineState(
...
+        pluginInfo.engineInfo,
+        walletInfo.keys.privateKeyFormat
+      ).filter(format => format !== privateKeyFormat)
 
-        const {
-          address: publicAddress,
-          nativeBalance = '0',
-          legacyAddress
-        } = await internalGetFreshAddress({
+      for (const format of supportedFormats) {
```
swansontec, line 379: I don't understand why this is a loop, if we only care about bip84. Would `if (supportedFormats.includes('bip84'))` do the same thing, since any other value seems to be a no-op?

swansontec (approved):
This looks good, but I do have one small question.

src/common/utxobased/engine/makeUtxoEngineState.ts
samholmes: You're quite right. I think I over-thought the code infrastructure here; I wrote as if any other supported addresses may be handled by the loop in the future, but there is no handlers for these cases anyway, so perhaps simple if statements instead a of a loop would be more straight-forward.

samholmes: I've made a change to no longer uselessly derive formats which don't impact the return value. I've left the loop in place as infrastructure to add support for other formats for the bip49 private key format (e.g. bech32m/taproot/bip350).

PR #361, opened 2023-02-03
Make edge-core-js a devDependency

swansontec:
The core should either be a peerDependency or a devDependency, since we don't need our own copy. We don't call core methods at runtime - the core calls us. We do need the types at build-time, so I have gone with devDependency.


swansontec:
Travis passed, but the GitHub build timed out, probably because of some network glitch.

samholmes (approved):

PR #386, opened 2023-08-25
Modern memos

swansontec:
### CHANGELOG

- changed: Add modern memo support via our currency info

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205362634032682

### DIFF
--- a/src/common/validateMemos.ts
+++ b/src/common/validateMemos.ts
@@ -1,0 +1,107 @@
+import { gt } from 'biggystring'
+import { asMaybe } from 'cleaners'
+import {
+  EdgeCurrencyInfo,
+  EdgeMemo,
+  EdgeMemoOption,
+  EdgeSpendInfo
+} from 'edge-core-js/types'
+
+import { asBase16, asIntegerString } from '../util/cleaners'
+
+/**
+ * Validates the memos on an spend request.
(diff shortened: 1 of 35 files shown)


src/common/validateMemos.ts
```diff
@@ -0,0 +1,107 @@
...
+    const value = asMaybe(asBase16)(memo.value)
+    if (value == null) {
+      return `is not valid hexadecimal`
+    }
+    if (option.maxBytes != null && value.length > option.maxBytes) {
+      return `cannot be longer than ${option.maxBytes} bytes`
+    }
+    if (option.minBytes != null && value.length < option.minBytes) {
+      return `cannot be shorter than ${option.minBytes} bytes`
+    }
```
samholmes, line 105: Multiply max/min by 2

samholmes (changes requested):
We'll need to implement support for using new memos field on the `EdgeTransaction` in the `makeSpend`.

src/common/validateMemos.ts
swansontec: Turns out, these have already been converted to UInt8Array's via `asBase16`, so the counts are already correct.

samholmes (approved):

PR #387, opened 2023-09-12
Enable AltcoinJS fee check using currency's `maximumFeeRate` config

samholmes:
### CHANGELOG

- changed: Re-enabled maximum fee rate checks from AltcoinJS.
- added: Configurable `maximumFeeRate` field for currency info's `EngineInfo` to change from default 5000 sats-per-vbyte

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205468832350747


paullinator (approved):

src/common/plugin/util/maximumFeeRateCalculator.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export const maximumFeeRateCalculator = (
+  currencyInfo: EdgeCurrencyInfo,
+  usdPrice: number
+): string | undefined => {
+  const { currencyCode, denominations } = currencyInfo
+  const primaryDenomination = denominations.find(
+    denom => denom.name === currencyCode
+  )
+  if (primaryDenomination == null) return undefined
```
paullinator, line 15: This should be a throw as must always have a denom for the primary currency

src/common/plugin/util/maximumFeeRateCalculator.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  const decimals: number = primaryDenomination.multiplier.length - 1
+  return Math.round((1 / usdPrice) * 10 ** decimals).toString()
```
paullinator, line 18: Just do the math with biggystring so you don't convert to and from strings/numbers.

This also ensures you never get scientific notation

```
import { div, mul, round } from 'biggystring

return round(mul(div('1', String(usdPrice), 16), primaryDenomination.multiplier))
```

paullinator (changes requested):

paullinator (approved):

PR #390, opened 2023-10-27
Add optional fallback generic parameter to asFeeInfo cleaner

samholmes:
In order to apply defaults to new entry additions to the FeeInfo type
(such as 'maximumFeeRate') without the fee info from disk/info-server
overwriting, we must apply the fallback values at the cleaner level.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205823029134351

### DIFF
--- a/src/common/plugin/types.ts
+++ b/src/common/plugin/types.ts
@@ -161,21 +161,28 @@ export type FeeInfo = FeeRates & {
   standardFeeHighAmount: string
   maximumFeeRate?: string
 }
-export const asFeeInfo = asObject<FeeInfo>({
-  ...asFeeRates.shape,
+export const asFeeInfo = (fallback?: FeeInfo): Cleaner<FeeInfo> =>
+  asObject<FeeInfo>({
+    ...asFeeRates.shape,
 
-  lowFeeFudgeFactor: asMaybe(asString),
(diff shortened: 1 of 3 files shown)


src/common/plugin/types.ts
```diff
@@ -161,21 +161,22 @@ export type FeeInfo = FeeRates & {
...
+    standardFeeLowFudgeFactor: asMaybe(asString),
+    standardFeeHighFudgeFactor: asMaybe(asString),
+    highFeeFudgeFactor: asMaybe(asString),
+
+    // The amount of satoshis which will be charged the standardFeeLow
+    standardFeeLowAmount: asString,
+    // The amount of satoshis which will be charged the standardFeeHigh
+    standardFeeHighAmount: asString,
+    // A safe-guard for any potential software bugs:
+    maximumFeeRate: asOptional(asString, fallback?.maximumFeeRate)
```
paullinator, line 178: Even though it's not really used yet, it seems like the fallback should be added to all the `asMaybe` parameters otherwise the `fallback` param is misleading as it really only applies to the `maximumFeeRate`

paullinator (approved):
Optional fix to make fallback live up to it's name

PR #391, opened 2023-11-08
Add missing check for currency RBF capabilities in accelerate method

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205888764853628

### DIFF
--- a/src/common/utxobased/engine/makeUtxoEngine.ts
+++ b/src/common/utxobased/engine/makeUtxoEngine.ts
@@ -141,6 +141,9 @@ export async function makeUtxoEngine(
 
   const engine: EdgeCurrencyEngine = {
     async accelerate(edgeTx: EdgeTransaction): Promise<EdgeTransaction | null> {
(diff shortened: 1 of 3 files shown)


src/common/utxobased/engine/makeUtxoEngine.ts
```diff
@@ -141,6 +141,8 @@ export async function makeUtxoEngine(
 
   const engine: EdgeCurrencyEngine = {
     async accelerate(edgeTx: EdgeTransaction): Promise<EdgeTransaction | null> {
+      if (currencyInfo.canReplaceByFee !== true) return null
```
swansontec, line 144: Although this way is longer, I would rather make the default `canReplaceByFee` value explicit:

```typescript
const { canReplaceByFee = false} = currencyInfo
if (!canReplaceByFee) return null
```

swansontec (approved):
Approved, with an optional change.

PR #395, opened 2024-01-26
Disable accelerate for bip44 wallets

peachbits:
---
  - https://app.asana.com/0/0/1206437518749021


samholmes (approved):
I figured this was needed when reviewing the last PR. I should have been more outspoken. All well

PR #396, opened 2024-04-03
Add NOWNodes to EngineInfo

samholmes:
This infra is similar to accountbased currencies in that we maintain
a list of server configs that can match with a specific key from
initOptions.
The only difference with this implementation is that there is no
info-server support for this integration yet.
These servers will only be used as fallback servers if WebSocket
broadcasts fail.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206574802387335


test/common/utxobased/network/Blockbook.spec.ts
```diff
@@ -151,7 +151,9 @@ describe('Blockbook', function () {
   const satoshiAddress = '1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa'
   const engineEmitter = new EngineEmitter()
   const socketEmitter = new SocketEmitter()
+
   const log = makeFakeLog()
+
   let blockBook: Blockbook
```
swansontec: Looks like you missed a camel-case `blockBook` here.

src/common/plugin/types.ts
```diff
@@ -79,6 +81,11 @@ export interface EngineInfo {
   ) => IProcessorTransaction
 }
 
+export interface ServerConfig {
+  type: 'blockbook-nownode'
+  uris: string[]
+}
```
swansontec, line 87: This structure seems weird. Why not put `fallbackNownodeServers?: string[]` straight into the engine info / init info?

I mean, it's always possible to invent your own custom object type using arrays, like `type FakeObject<T> = Array<{ key: string, value: T}>`, but this sort of thing is rarely a good idea. Just use normal objects.

The `ServerConfig` type seems to be implementing this anti-pattern, with `type` as the key and `uris` as the value.

src/common/utxobased/engine/ServerStates.ts
```diff
@@ -304,38 +320,117 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
 
   const broadcastTx = async (transaction: EdgeTransaction): Promise<string> => {
     return await new Promise((resolve, reject) => {
```
swansontec, line 322: Since you are refactoring this guy pretty heavily, could we fix the strange `new Promise` thing into normal async / await? The structure seems pretty linear, so it should be a straightforward conversion.

OTOH, what you have works, even if it weird to read.

swansontec (approved):
I am approving this because it seems like it would work fine as-is.

There are some things that could probably be improved, though, so I left comments for those.

src/common/plugin/types.ts
samholmes: Because: It's easier to enumerate nominal types then fields in typescript.

src/common/utxobased/engine/ServerStates.ts
samholmes: I was going to, but it would change the behavior from parallel promises to sequential which is not what is expected for broadcast. I'm just honoring the implementation; but if you have a good argument for sequentializing it then I can consider.

swansontec: Right, which is why we normally use `Promise.all` in cases like this. It's fine to leave it as-is, though.

src/common/plugin/types.ts
swansontec: Ok, we discussed this design some more, and it makes sense if the servers need extra data, such as contract addresses or such. In that case, it might be `{ type: 'foo', servers: [], queryParams: '' }`, and so we want the flexibility of the array approach. This will impact the way the info server assembles the core rollup payload, so we might need a follow-up PR over there.

PR #397, opened 2024-04-18
Currency info cleanup

swansontec:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Description

It turns out we already had fee templates in here, so I just reorganized things to match edge-currency-accountbased a bit more.

---
  - https://app.asana.com/0/0/1206994839543313

### DIFF
--- a/src/common/utxobased/info/litecoin.ts
+++ b/src/common/utxobased/info/litecoin.ts
@@ -3,12 +3,18 @@ import { EdgeCurrencyInfo } from 'edge-core-js/types'
 import { IMAGE_SERVER_URL } from '../../constants'
 import { CoinInfo, EngineInfo, PluginInfo } from '../../plugin/types'
 import { maximumFeeRateCalculator } from '../../plugin/util/maximumFeeRateCalculator'
-import { memoInfo } from './commonInfo'
+import {
+  legacyMemoInfo,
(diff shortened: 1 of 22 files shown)


src/common/utxobased/info/litecoin.ts
```diff
@@ -6,10 +6,12 @@ import { maximumFeeRateCalculator } from '../../plugin/util/maximumFeeRateCalcul
 import { memoInfo } from './commonInfo'
 
 export const currencyInfo: EdgeCurrencyInfo = {
-  pluginId: 'litecoin',
-  walletType: 'wallet:litecoin',
+  canReplaceByFee: false,
```
samholmes, line 13: I feel like `canReplaceByFee` field is in a different "group" or category of fields for the EdgeCurrencyInfo. Perhaps it would make sense to put a comment above `canReplaceByFee` and potentially other fields like it and separate it from the required fields like `pluginId` and `currencyCode`.

samholmes: Seeing the following commits after this comment made me realize the change request is too much rebasing effort to be worth it. I'll leave the idea here as optional though.

samholmes (approved):

PR #399, opened 2024-04-24
Upgrade core

swansontec:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207134573674963

### DIFF
--- a/src/common/utxobased/db/Processor.ts
+++ b/src/common/utxobased/db/Processor.ts
@@ -33,7 +33,12 @@ interface FetchTransactionArgs {
   blockHeight?: number
   blockHeightMax?: number
   txId?: string
-  options?: EdgeGetTransactionsOptions
+  options?: EdgeGetTransactionsOptions & {
+    endDate?: Date
+    startDate?: Date
+    startEntries?: number
+    startIndex?: number
+  }
 }
 
 interface FetchUtxosArgs {
--- a/src/common/utxobased/engine/types.ts
+++ b/src/common/utxobased/engine/types.ts
@@ -42,13 +42,15 @@ export const asUtxoSignMessageOtherParams = asObject({
   publicAddress: asString
 })
 
+const asOutputSort = asValue('bip69', 'targets')
+
 export type UtxoSpendInfoOtherParams = ReturnType<
   typeof asUtxoSpendInfoOtherParams
 >
 export const asUtxoSpendInfoOtherParams = asObject({
   enableRbf: asOptional(asBoolean),
   forceChangeAddress: asOptional(asString),
-  outputSort: asOptional(asValue('bip69', 'targets'), 'bip69'),
+  outputSort: asOptional(asOutputSort, 'bip69'),
   txOptions: asOptional(asTxOptions),
   utxoSourceAddress: asOptional(asString)
 })
--- a/src/common/utxobased/info/bitcoin.ts
+++ b/src/common/utxobased/info/bitcoin.ts
@@ -1,6 +1,5 @@
 import { EdgeCurrencyInfo } from 'edge-core-js/types'
 
-import { IMAGE_SERVER_URL } from '../../constants'
 import { CoinInfo, EngineInfo, PluginInfo } from '../../plugin/types'
 import { maximumFeeRateCalculator } from '../../plugin/util/maximumFeeRateCalculator'
 import {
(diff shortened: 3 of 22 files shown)


src/common/utxobased/info/bitcoin.ts
```diff
@@ -23,10 +22,6 @@ const currencyInfo: EdgeCurrencyInfo = {
   addressExplorer: 'https://blockchair.com/bitcoin/address/%s',
   transactionExplorer: 'https://blockchair.com/bitcoin/transaction/%s',
 
-  // Images:
-  symbolImage: `${IMAGE_SERVER_URL}/bitcoin-logo-solo-64.png`,
-  symbolImageDarkMono: `${IMAGE_SERVER_URL}/bitcoin-logo-solo-64.png`,
```
samholmes, line 28: Ah nice, I just checked that this isn't used in the GUI either.

src/common/utxobased/db/Processor.ts
```diff
@@ -368,28 +368,12 @@ export async function makeProcessor(
...
-              '',
-              unixTime(startDate.getTime()),
-              unixTime(endDate.getTime())
-            )
-          }
+          const txData: TxIdByDate[] = await tables.txIdsByDate.query(
+            '',
+            0,
+            unixTime(Date.now())
+          )
```
samholmes, line 376: How is this change relevant to the upgrade I wonder?

test/common/utxobased/db/Processor.spec.ts
```diff
@@ -589,21 +589,10 @@ describe('Processor transactions tests', () => {
...
-    })
-    expect(results).to.deep.equal([])
-
-    const [tx6] = await processor.fetchTransactions({
-      options: {
-        startDate: new Date(9_000),
-        endDate: new Date(15_000)
-      }
+    const [tx4] = await processor.fetchTransactions({
+      options: { tokenId: null }
```
samholmes, line 593: Are we deprecating some internal API without specifying the deprecation?

test/common/utxobased/engine/engine.spec.ts
```diff
@@ -286,30 +286,21 @@ describe('engine.spec', function () {
...
       })
-
-      it('Should get transactions from cache with options', async function () {
-        const options: EdgeGetTransactionsOptions = {
-          startIndex: 1,
-          startEntries: 2
-        }
-        const txs = await engine.getTransactions(options)
-        assert.equal(txs.length, 2, 'should have 2 tx from cache')
-      })
```
samholmes, line 310: Why was this test removed?

samholmes (changes requested):
More questions again

test/common/utxobased/db/Processor.spec.ts
swansontec: The core removed these features, and the removals were all documented in the release notes. Specifically, we removed all the options to `getTransactions` besides the `tokenId`, so the core is in charge of handling the pagination, searching, and other advanced query options.

test/common/utxobased/engine/engine.spec.ts
swansontec: See above. The core removed the ability to pass any `getTransactions` flags besides the `tokenId`. Since we don't need these features, we don't need to test them.

src/common/utxobased/engine/types.ts
```diff
@@ -42,13 +42,15 @@ export const asUtxoSignMessageOtherParams = asObject({
...
+const asOutputSort = asValue('bip69', 'targets')
+
 export type UtxoSpendInfoOtherParams = ReturnType<
   typeof asUtxoSpendInfoOtherParams
 >
 export const asUtxoSpendInfoOtherParams = asObject({
   enableRbf: asOptional(asBoolean),
   forceChangeAddress: asOptional(asString),
-  outputSort: asOptional(asValue('bip69', 'targets'), 'bip69'),
+  outputSort: asOptional(asOutputSort, 'bip69'),
```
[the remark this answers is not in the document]
swansontec: Added the message: "When the asValue cleaner gets embedded inside asObject, it will often infer string instead of the specific literal type. Moving the asValue cleaner to its own line allows TypeScript to infer the exact string literals, instead of the string catch-all."

test/common/utxobased/db/Processor.spec.ts
samholmes: Ah, cause the processor coupled its type with the core originally, the core upgrade impacts the processor API. Got it.

src/common/utxobased/db/Processor.ts
samholmes: My final question now is: is there a new protocol for the core to query the plugin for paging or streaming of this data that we should mark as a TODO?

samholmes (approved):
Approved, but with one final question or cosideration.

PR #402, opened 2024-05-10
Webpack into native module

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207103682121388

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,62 @@
-# edge-currency-plugins 
+# Edge Currency Plugin for UTXO-Based currencies
 
-> Plugins for edge-core-js to support various cryptocurrencies such as bitcoin, bitcoin cash, and litecoin, 
+> Plugins for [edge-core-js](https://github.com/EdgeApp/edge-core-js), to support various cryptocurrencies such as bitcoin, bitcoin cash, and litecoin,
+
+## Installing
+
+First, add this library to your project:
+
+```sh
+npm i -s edge-currency-plugins
+```
+
+### Node.js
+
+For Node.js, you should call `addEdgeCorePlugins` to register these plugins with edge-core-js:
+
+```js
+const { addEdgeCorePlugins, lockEdgeCorePlugins } = require("edge-core-js");
+const plugins = require("edge-currency-plugins");
+
+addEdgeCorePlugins(plugins);
+
+// Once you are done adding plugins, call this:
+lockEdgeCorePlugins();
+```
+
+You can also add plugins individually if you want to be more picky:
+
+```js
+addEdgeCorePlugins({
+  bitcoin: plugins.bitcoin,
+});
+```
+
+### Browser
+
+The bundle located in `dist/edge-currency-plugins.js` will automatically register itself with edge-core-js. Just serve the entire `dist` directory along with your app, and then load the script:
+
+```html
+<script src='https://example.com/app/dist/edge-currency-plugins.js'>
+```
+
+If you want to debug this project, run `yarn start` to start a Webpack server,
+and then adjust your script URL to http://localhost:8081/edge-currency-plugins.js.
+
+### React Native
+
+The `edge-currency-plugins` package will automatically install itself using React Native autolinking.
+To integrate the plugins with edge-core-js, add its URI to the context component:
+
+```jsx
+import { pluginUri, makePluginIo } from "edge-currency-plugins";
+<MakeEdgeContext
+  nativeIo={{
+    "edge-currency-plugins": makePluginIo(),
+  }}
+  pluginUris={[pluginUri]}
+  // Plus other props as required...
+/>;
+```
+
+To debug this project, run `yarn start` to start a Webpack server, and then use `debugUri` instead of `pluginUri`.
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,0 +1,45 @@
+buildscript {
+  repositories {
+    google()
+    mavenCentral()
+  }
+
+  dependencies {
+    classpath 'com.android.tools.build:gradle:3.6.0'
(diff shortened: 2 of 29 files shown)


android/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+def safeExtGet(prop, fallback) {
+  rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
+
+def DEFAULT_COMPILE_SDK_VERSION = 28
+def DEFAULT_BUILD_TOOLS_VERSION = '28.0.2'
+def DEFAULT_MIN_SDK_VERSION = 19
+def DEFAULT_TARGET_SDK_VERSION = 27
+
+android {
```
swansontec, line 23: Please add the package name in here: ` namespace "app.edge.reactnative.currencyplugins"`, and then remove it from the Android Manifest.

Putting the namespace in the manifest works, but is deprecated.

android/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,3 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          package="app.edge.reactnative.currencyplugins">
```
swansontec, line 2: We can remove the `package=` prop once we add it to the build.grade.

webpack.config.js
```diff
@@ -0,0 +1,84 @@
...
+const webpack = require('webpack')
+
+// Run `yarn start.plugins` to enable debug mode.
+// This mode will serve the plugin bundle via a local dev-server.
+const debug = process.env.WEBPACK_SERVE
+
+// Try exposing our socket to adb (errors are fine):
+if (process.env.WEBPACK_SERVE) {
+  console.log('adb reverse tcp:8101 tcp:8101')
+  exec('adb reverse tcp:8101 tcp:8101', () => {})
```
swansontec, line 14: This port, 8101, the the one the GUI currently uses. That is fine!

However, edge-core-js uses 8080, edge-currency-accountbased used 8082, and edge-exchange plugins uses 8083. So it seems like 8081 would "fit" better than 8101.

webpack.config.js
```diff
@@ -0,0 +1,84 @@
+const { exec } = require('child_process')
+const path = require('path')
+const { EsbuildPlugin } = require('esbuild-loader')
+
+const webpack = require('webpack')
+
+// Run `yarn start.plugins` to enable debug mode.
+// This mode will serve the plugin bundle via a local dev-server.
```
swansontec, line 8: This yarn instruction is incorrect.

webpack.config.js
```diff
@@ -0,0 +1,84 @@
...
+      https: require.resolve('https-browserify'),
+      os: require.resolve('os-browserify/browser'),
+      path: require.resolve('path-browserify'),
+      stream: require.resolve('stream-browserify'),
+      string_decoder: require.resolve('string_decoder'),
+      url: require.resolve('url'),
+      vm: require.resolve('vm-browserify')
+    },
+    mainFields: ['browser', 'module', 'main']
+  }
```
swansontec, line 83: Now that this config file only needs to worry about edge-currency-plugins and not anybody else, I'm guessing we can really minimize this section. Does edge-currency-plugins really depend on all this Node.js stuff? Probably not all of it.

However, I understand the motivation to keep things stable for now. Maybe we can create a follow-on task and handle this cleanup in a future release.

README.md
```diff
@@ -1,3 +1,62 @@
...
+
+// Once you are done adding plugins, call this:
+lockEdgeCorePlugins();
+```
+
+You can also add plugins individually if you want to be more picky:
+
+```js
+addEdgeCorePlugins({
+  ethereum: plugins.ethereum,
```
swansontec, line 31: Probably a bad example for this repo.

README.md
```diff
@@ -1,3 +1,62 @@
...
+### Browser
+
+The bundle located in `dist/edge-currency-plugins.js` will automatically register itself with edge-core-js. Just serve the entire `dist` directory along with your app, and then load the script:
+
+```html
+<script src='https://example.com/app/dist/edge-currency-plugins.js'>
+```
+
+If you want to debug this project, run `yarn start` to start a Webpack server,
+and then adjust your script URL to http://localhost:8101/edge-currency-plugins.js.
```
swansontec, line 44: If we do adjust the port number, change it here too.

src/react-native.ts
```diff
@@ -0,0 +1,13 @@
+import { EdgeOtherMethods } from 'edge-core-js/types'
+import { NativeModules } from 'react-native'
+// import { bridgifyObject, emit, onMethod } from 'yaob'
+
+const { EdgeCurrencyPluginsModule } = NativeModules
+const { sourceUri } = EdgeCurrencyPluginsModule.getConstants()
+
+export const pluginUri = sourceUri
+export const debugUri = 'http://localhost:8101/edge-currency-plugins.js'
```
swansontec, line 8: If we do change the port number, adjust it here too.

swansontec (changes requested):
Just a few small things need to get fixed.

android/src/main/AndroidManifest.xml
samholmes: Hmm, this is not how it is configured in accountbased.

android/build.gradle
samholmes: Hmm, this is not how it is configured in accountbased.

swansontec: See edge-core-js for an example. I guess we need to update accountbased too.

swansontec (approved):

PR #406, opened 2024-06-07
Use new `enableRbf` field on EdgeSpendInfo

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207517164181978

### DIFF
--- a/src/common/utxobased/engine/UtxoEngine.ts
+++ b/src/common/utxobased/engine/UtxoEngine.ts
@@ -617,6 +617,13 @@ export async function makeUtxoEngine(
         targets = []
       }
 
+      // Determine whether the transaction should enable RBF given that the
+      // currency supports RBF.
+      const enableRbf =
+        currencyInfo.canReplaceByFee === true
+          ? edgeSpendInfo.enableRbf ?? spendInfoOtherParams.enableRbf ?? true
+          : false
+
       log.warn(`spend: Using fee rate ${feeRate} sat/B`)
       const subtractFee =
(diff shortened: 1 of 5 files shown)


src/common/utxobased/engine/types.ts
```diff
@@ -48,7 +48,6 @@ export type UtxoSpendInfoOtherParams = ReturnType<
   typeof asUtxoSpendInfoOtherParams
 >
 export const asUtxoSpendInfoOtherParams = asObject({
-  enableRbf: asOptional(asBoolean),
```
swansontec, line 51: This is a breaking change. You should leave it in with a `/** @deprecated */` flag.

src/common/utxobased/engine/UtxoEngine.ts
```diff
@@ -617,6 +617,13 @@ export async function makeUtxoEngine(
         targets = []
       }
 
+      // Determine whether the transaction should enable RBF given that the
+      // currency supports RBF.
+      const enableRbf =
```
swansontec, line 622: `const enableRbf = currencyInfo.canReplaceByFee === true ? edgeSpendInfo.enableRbf ?? spendInfoOtherParams.enableRbf ?? true : false` to preserve legacy behavior when the new flag is missing.

swansontec (approved):
Approved, but this is a breaking change. To make this a minor version bump, just some small tweaks are needed.

swansontec (approved):

PR #408, opened 2024-07-17
Plugin State Fixes

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207816094830396

### DIFF
--- a/src/common/plugin/PluginState.ts
+++ b/src/common/plugin/PluginState.ts
@@ -77,10 +77,16 @@ export function makePluginState(settings: PluginStateSettings): PluginState {
     internalServers: {}
   }
   let serverCacheDirty = false
-  let knownServers: ServerList = {}
+
+  const getSelectedServerList = (): ServerList => {
+    const serverCacheIndex = serverCache.enableCustomServers
+      ? 'customServers'
+      : 'internalServers'
+    return serverCache[serverCacheIndex]
+  }
 
   const saveServerCache = async (): Promise<void> => {
-    serverScores.printServers(knownServers)
+    serverScores.printServers(getSelectedServerList())
     if (serverCacheDirty) {
       await memlet.setJson(SERVER_CACHE_FILE, serverCache).catch(e => {
         log(`${pluginId} - ${JSON.stringify(e.toString())}`)
@@ -150,7 +156,7 @@ export function makePluginState(settings: PluginStateSettings): PluginState {
 
     dumpData(): JsonObject {
       return {
-        'pluginState.servers_': knownServers
+        'pluginState.servers_': getSelectedServerList()
       }
     },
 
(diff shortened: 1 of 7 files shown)


src/common/utxobased/network/Socket.ts
```diff
@@ -146,8 +146,8 @@ export function makeSocket(uri: string, config: SocketConfig): Socket {
   }
 
   const onSocketClose = (): void => {
-    const err = error ?? new Error('Socket close')
-    log.warn(`onSocketClose due to ${err.message} with server ${uri}`)
+    const err = error ?? new Error('Socket closed without error')
+    log.warn(`onSocketClose with server ${uri}: ${err.message}`)
```
swansontec: It is never valid to access `error.message` unless we first determine that the error is `instanceof Error`. Your `handleError` signature up above is wrong, since it should be `e: unknown` for the parameter type, and then follow the TypeScript squiggles until you've handled all the cases.

It might make sense to rename `let error: Error | undefined` to `let errorMessage: string | undefined`, and then `handleError` would have code like `errorMessage = e instanceof Error ? e.message : String(e)`. Then `onSocketClose` wouldn't need to call `new Error`, since we are dealing with plain strings now.

Edit: Oh, derp. You end up doing something similar over the next few commits. I still think you should do the `string` thing, and maybe smoosh the error-handling commits together into one.

src/common/plugin/PluginState.ts
```diff
@@ -219,9 +221,31 @@ export function makePluginState(settings: PluginStateSettings): PluginState {
     },
 
     async updateServers(settings: UtxoUserSettings): Promise<void> {
-      enableCustomServers = settings.enableCustomServers
-      customServers = settings.blockbookServers
+      if (settings === userSettings) return
+
+      const hasServerListChanged = (() => {
```
swansontec, line 234: Can this be a normal function, and then just invoke it in the `if` statement as `if (!hasServerListChanged() && ...`?

IIFE's are always confusing to read, because they look a lot like function definitions when scanning the code. In this case, it seems like an actual function definition would have worked just as well.

src/common/plugin/PluginState.ts
```diff
@@ -222,11 +229,11 @@ export function makePluginState(settings: PluginStateSettings): PluginState {
     },
 
     async updateServers(settings: UtxoUserSettings): Promise<void> {
-      if (settings === userSettings) return
```
swansontec: So we added this, and then removed it? Maybe squash the commits to avoid the detour?

swansontec (changes requested):
Found a few things.

src/common/utxobased/network/Socket.ts
samholmes: I think keeping the internal state as an `unknown` time is useful for debugging what gets thrown by a breakpoint in `onSocketClose`. Making `onSocketClose` the top level handler of the thrown thing will make it easier to figure out what weird thing happens at a lower level. For example, if we see `'[object Object]'` errors, then we know some weird object is thrown that cannot be converted to a string. And we can debug it by going to this one place in `onSocketClose` to figure out what the object is, instead of trying to search for where the throw occurred.

src/common/plugin/PluginState.ts
samholmes: It was relevant for the previous change but not for this one. This was one of the reason why I wanted to perhaps have a second PR for these changes since this is really a bug fix on top of the other fix... 

I can actually remove the previous line since it's just an early exit/optimization on the previous fix. This would make the PR more cohesive.

samholmes: Note the last commit change a little by effect of this: https://github.com/EdgeApp/edge-currency-plugins/pull/408/commits/1f3e34641442a5d6f4953326c7a34635241b3173#diff-e333c8da8576d31b705ec8d0d92bea632b797f90b76f50b2c961371dd8570b2fR233

I move the constant closer to its place of use.

swansontec (approved):

src/common/utxobased/network/Socket.ts
swansontec: That makes sense.

PR #411, opened 2024-08-14
Add NOWNodes Blockbook servers

samholmes:
### Note to reviewer

Please suggest a better way to handle the score lowering for URIs that contain "key params".

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none


src/util/uriKeyParams.ts
```diff
@@ -0,0 +1,20 @@
+export function getUriKeyParams(uri: string): string[] {
+  const params = uri.match(/%\{(\w|\d)+\}/g)
+  return params == null ? [] : params
+}
+
+export function replaceKeyParams<T extends string>(
+  uri: string,
+  options: { [P in T]?: string }
+): string {
+  const keyParams = getUriKeyParams(uri)
```
swansontec, line 10: The `getUriKeyParams` should just be inlined as `const keyParams = uri.match(/%\{(\w|\d)+\}/g) ?? []`

swansontec (approved):
One cleanup would be nice, but otherwise it seems like this will work fine.

PR #414, opened 2024-09-23
Force enableCustomServers to false server list is empty

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207859097985661

### DIFF
--- a/src/common/plugin/PluginState.ts
+++ b/src/common/plugin/PluginState.ts
@@ -248,6 +248,20 @@ export function makePluginState(settings: PluginStateSettings): PluginState {
         return
       }
 
+      // Force enableCustomServers to false server list is empty because
+      // an empty server list would cause the wallets to never sync, and
+      // this is unlikely the user's intention.
+      // This means policy change was decided later, which makes
+      // `enableCustomServers` functionally only useful for disabling custom
+      // servers entirely even when passing a list of servers (which could have
+      // been done by passing an empty list to begin with).
+      // In other words, this field is more of an internal field to track whether
+      // to treat the list a static or not (whether to fetch serve lists outside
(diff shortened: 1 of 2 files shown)


src/common/plugin/PluginState.ts
```diff
@@ -248,6 +248,20 @@ export function makePluginState(settings: PluginStateSettings): PluginState {
...
+      // This means policy change was decided later, which makes
+      // `enableCustomServers` functionally only useful for disabling custom
+      // servers entirely even when passing a list of servers (which could have
+      // been done by passing an empty list to begin with).
+      // In other words, this field is more of an internal field to track whether
+      // to treat the list a static or not (whether to fetch serve lists outside
+      // of the list provided by the user or not).
+      const enableCustomServers =
+        newSettings.enableCustomServers &&
+        newSettings.blockbookServers.length === 0
```
swansontec, line 263: I can't comprehend this logic. The `&&` is really tripping me up. I fee like It would make sense if the second term was `!== 0`, but this way seems backwards. Has this been tested? What am I missing here?

swansontec (approved):
I am approving this, because even if my concern is correct, it's a one-character change. If my concern is not correct, and I'm just misunderstanding the code, then it's all good to go.

src/common/plugin/PluginState.ts
samholmes: I believe this is a valid catch. I had originally had it as a tertiary operator and simplified to a conjunction and missed that I needed to invert the condition from `===` to `!==`.

swansontec (approved):

PR #419, opened 2024-12-19
Implement updateInfoPayload

samholmes:
- **Fix broken cleaner for atomicwallet.io**
- **Implement updateInfoPayload**
- **Simplify options for EngineConfig**

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208931449878115

### DIFF
--- a/src/common/plugin/CurrencyPlugin.ts
+++ b/src/common/plugin/CurrencyPlugin.ts
@@ -1,3 +1,4 @@
+import { asMaybe } from 'cleaners'
 import {
   EdgeCorePluginOptions,
   EdgeCurrencyEngine,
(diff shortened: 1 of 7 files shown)


src/common/plugin/CurrencyPlugin.ts
```diff
@@ -81,6 +86,16 @@ export function makeCurrencyPlugin(
...
       })
       return currencyTools
+    },
+
+    async updateInfoPayload(infoPayload) {
+      try {
+        cachedInfoPayload = asInfoPayload(infoPayload)
+      } catch (_) {
+        log.warn('invalid infoPayload', infoPayload)
+      }
```
swansontec, line 95: The plugin state object has a member called `getInfoPayload`, but this seems weirdly indirect. The plugin state object could just have an `infoPayload` member, which is the value itself.

Then here we can do `pluginState.infoPayload = asInfoPayload(infoPayload)` to update the value directly, and in the constructor we can just pass `infoPayload` as a value. Fixing this would remove a level of indirection.

However, I understand that this is just a style thing, so if you disagree, feel free to ignore. The way you have it certainly works.

swansontec (approved):

PR #420, opened 2025-01-02
Integrate Seen Tx Checkpoint API

samholmes:
- **Change Blockbook connection for test**
- **Upgrade edge-core-js@^2.22.1**
- **Remove `transactionChanged`**
- **Integrate new Seen Tx Checkpoint API**

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208861595781414

### DIFF
--- a/src/common/utxobased/engine/UtxoEngine.ts
+++ b/src/common/utxobased/engine/UtxoEngine.ts
@@ -61,10 +61,7 @@ import {
   pathToPurposeType,
   sumUtxos
 } from './utils'
-import {
-  makeUtxoEngineProcessor,
-  transactionChanged
-} from './UtxoEngineProcessor'
+import { makeUtxoEngineProcessor } from './UtxoEngineProcessor'
 import { makeUtxoWalletTools } from './UtxoWalletTools'
 
 export async function makeUtxoEngine(
@@ -133,10 +130,10 @@ export async function makeUtxoEngine(
 
   const engineProcessor = makeUtxoEngineProcessor({
     ...config,
-    walletTools,
-    walletInfo,
     dataLayer,
-    pluginState
+    pluginState,
+    walletTools,
+    walletInfo
   })
 
   const engine: EdgeCurrencyEngine = {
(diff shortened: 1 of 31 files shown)


src/common/utxobased/engine/UtxoEngineProcessor.ts
```diff
@@ -654,7 +683,7 @@ interface BlockbookUtxoCache {
 interface TransactionSpecificUpdateCache {
   [key: string]: { processing: boolean }
 }
-interface TransactionUpdateCache {
+interface transactionUpdateCache {
```
swansontec, line 686: Interface names should always be uppercase.

test/common/utxobased/engine/engine.spec.ts
```diff
@@ -335,7 +340,7 @@ describe('engine.spec', function () {
               done() // Can be "done" since the promise resolves before the event fires but just be on the safe side
             }
           })
-          engine.startEngine().catch(e => {
+          engine.startEngine({ seenTxCheckpoint: '' }).catch(e => {
```
swansontec, line 343: Shouldn't this be undefined, instead of a blank string?

If the core types don't allow that, then we need to fix the core.

swansontec (changes requested):
Only small changes needed here.

src/common/utxobased/engine/UtxoEngine.ts
```diff
@@ -128,10 +128,24 @@ export async function makeUtxoEngine(
   // private keys.
   let nonceDataLayer: DataLayer | undefined
 
+  // This cached value allows the engine to resync using the same checkpoint as
+  // what the core gave the plugin when the core started plugin.
```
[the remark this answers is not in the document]
samholmes: The suggested comment isn't the same. What I'm trying to convey in the comment is that this is a cached value for the plugin (a stateful function), and it is needed to be cached as the same type as what was received from the core so that way it can be re-used in `resyncBlockchain`.

The core treats the value as a string and the plugin derives another type from this string. The plugin will emit the string back to the core via `onSeenTxCheckpoint` at a later point. So, it needs to keep the original string intact for that callback to the core. This is why I chose to keep the string as state and let the derived state be ephemeral.

swansontec (approved):

PR #422, opened 2025-02-04
Fix checkpoint initialization on upgrade

swansontec:
We were only updating the checkpoint in response to incoming transactions, which doesn't happen if the phone is already synced. The next incoming transaction after the upgrade will encounter and `undefined` checkpoint, and won't generate a notifiation. Instead, also update the checkpoint when the sync finishes, to ensure that existing accounts will detect incoming funds without needing a full resync.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209313363246606

### DIFF
--- a/src/common/utxobased/engine/UtxoEngineProcessor.ts
+++ b/src/common/utxobased/engine/UtxoEngineProcessor.ts
@@ -168,6 +168,7 @@ export function makeUtxoEngineProcessor(
         `processed changed, percent: ${percent}, processedCount: ${processedCount}, totalCount: ${expectedProcessCount}`
       )
       processedPercent = percent
+      common.updateSeenTxCheckpoint()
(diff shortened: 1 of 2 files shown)


src/common/utxobased/engine/UtxoEngineProcessor.ts
```diff
@@ -168,6 +168,7 @@ export function makeUtxoEngineProcessor(
         `processed changed, percent: ${percent}, processedCount: ${processedCount}, totalCount: ${expectedProcessCount}`
       )
       processedPercent = percent
+      common.updateSeenTxCheckpoint()
```
samholmes, line 171: Ah, yes this is what I missed for currency plugins.

samholmes (approved):

PR #428, opened 2025-04-08
Replace secp256k1 library with bigint compatible version

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209576495542847

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "*.{js,ts,json}": "eslint"
   },
   "dependencies": {
-    "@bitcoin-js/tiny-secp256k1-asmjs": "^2.2.3",
+    "@bitcoinerlab/secp256k1": "^1.2.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -40,7 +40,7 @@
     "*.{js,ts,json}": "eslint"
   },
   "dependencies": {
-    "@bitcoin-js/tiny-secp256k1-asmjs": "^2.2.3",
+    "@bitcoinerlab/secp256k1": "^1.2.0",
```
samholmes, line 43: I wonder if https://github.com/bitcoinjs/tiny-secp256k1 would make more sense. It may be a breaking change because it requires WASM, but how many platforms that support BigInt don't support WASM?

samholmes (approved):

PR #454, opened 2026-07-08
Encode BIP137 SegWit message signature headers

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

[Asana task](https://app.asana.com/0/0/1216403654258303/f)

Edge's BTC message signing (used for CEX/ramp withdrawal ownership proofs, e.g. Bringin / NiceHash Travel Rule) produced a signature whose header byte always encoded a legacy P2PKH type (27-34), regardless of the signing address. BIP137 verifiers reject a legacy-header signature that comes from a SegWit address: nested SegWit (P2SH-P2WPKH) requires header 35-38 and native SegWit (P2WPKH / bech32) requires 39-42. Because Edge BTC wallets default to SegWit, the signatures Edge emitted were rejected by Bringin; a user had to run a script to rewrite the header to BIP137 before it was accepted, or fall back to selfie/ID verification.

`signMessageBase64` in `keymanager.ts` called `bitcoinMessage.sign(message, privKey, compressed)` with no `segwitType`, so the header byte only reflected the compressed flag. This threads the address's derivation `format` through and selects the matching `segwitType`:

- `bip84` (native SegWit / P2WPKH) -> header 39-42
- `bip49` (nested SegWit / P2SH-P2WPKH) -> header 35-38
- `bip44` / `bip32` (legacy P2PKH) -> header 31-34 (unchanged)

The header selection routes through the canonical `BIP43PurposeTypeEnum` rather than raw format string literals, so a newly-added SegWit format fails loudly in the purpose mapping instead of silently falling through to a legacy header.

The ECDSA signature is unchanged for legacy P2PKH signing on Bitcoin; only the header byte differs there. The fix is general across all UTXO coins that support SegWit, since `path.format` reflects the actual address derivation.

**BIP137 is opt-in**

`asUtxoSignMessageOtherParams` carries `signatureFormat: 'electrum' | 'bip137'`, defaulting to `electrum`. `segwitType` is applied only when `bip137` is requested, so existing callers (`bityProvider`, `EdgeProviderServer`) keep the header byte they produce today. Emitting BIP137 unconditionally would have been a silent behavior change for them, and it would have broken [EdgeApp/edge-react-gui#6065](https://github.com/EdgeApp/edge-react-gui/pull/6065) depending on merge order: that PR's "Standard (Electrum)" option would have started emitting BIP137 headers. Making it opt-in removes the ordering hazard, and the GUI now sends the option instead of rewriting the header itself.

**Scoped purpose lookup**

`bip43PurposeNumberToTypeEnum` throws on a purpose it does not map, so resolving it unconditionally would let a future `CurrencyFormat` (bip86 Taproot being the obvious candidate) break message signing on the default Electrum path, which never needs it. The lookup lives in a `getBip137SegwitType` helper that returns before touching it unless BIP137 is requested, so an unmapped format can only affect BIP137 signing.

**`AddressNotOwnedError`**

`signMessage` throws a named `AddressNotOwnedError` when the wallet cannot sign for the requested address, whether it fails to parse or simply is not ours. Callers branch on `error.name` rather than matching parser prose, following the existing `asMaybeInsufficientFundsErrorPlus` pattern.

**Coin message prefixes**

The header byte alone is not enough for non-Bitcoin coins. `signMessageBase64` previously magic-hashed with `bitcoinjs-message`'s default `"Bitcoin Signed Message:\n"` for every coin, so an altcoin signature carried the right header but the wrong prefix and would not verify against that coin's own addresses. It now hashes with the coin's own `prefixes.messagePrefix[0]`.

That made the prefix data load-bearing for the first time, and an audit by @peachbits (commit absorbed here, authored by them and ordered first) found 14 coins wrong. `bitcoinjs-message` copies the prefix into the hash verbatim and only varint-encodes the message length, so the prefix must carry its own leading CompactSize byte, mirroring `ss << strMessageMagic` in each coin's C++. Two failure classes: a wrong length byte (litecoin, dogecoin, feathercoin, zcoin, bitcoingold and its testnet) and a wrong string, usually Bitcoin's default left in place (dash, pivx, ecash, qtum, ravencoin, smartcash, ufo, digibyte). Dash's `DarkCoin`, PIVX's `DarkNet`, and Vertcoin's genuine use of Bitcoin's magic are pre-rebrand or intentional and are commented so they do not get "fixed" back.

**Testing**

- Unit tests pin both encodings on bitcoin and litecoin: the default produces the same `H…` header for `bip44`, `bip49` and `bip84`, and only an explicit `bip137` request shifts `bip49` to `I…` and `bip84` to `J…`. 12 cases, so swapping the two formats fails the suite. `tsc`, `eslint`, and the full `mocha` suite pass (1251 tests); `npm run verify` passes.
- `messagePrefix.spec.ts` asserts every registered coin's prefix leading byte matches its string length, covering 24 coins. This is the guard that makes a silently-unverifiable prefix a build failure.
- Litecoin sign-message vectors verify via `bitcoinjs-message.verify` against the derived legacy, P2SH-P2WPKH, and bech32 LTC addresses, and are confirmed NOT to verify under Bitcoin's prefix, so a regression back to the default would fail the suite.
- Offline round-trip: each Bitcoin format's signature verifies via `bitcoinjs-message.verify` against its real P2PKH / P2SH-P2WPKH / bech32 address.
- In-app (iOS sim, edge-funds BTC wallet): drove the real `EdgeCurrencyWallet.signMessage` path with this build linked, on the wallet's own receive addresses. The nested-SegWit `3...` address produced header 35 and the native-SegWit `bc1q...` address produced header 40, both verifying against their respective addresses (screenshot attached). Before the fix both emitted header 31, which fails BIP137 verification for a SegWit address.
- In-app, opt-in round (iOS sim, this build linked into edge-react-gui#6065 via `updot`): signing the same message with bip84 address `bc1q7f5wkp5k0847utsec7p8v59cqj7et6wrzz7dv4` produced header **32** under the default and header **40** under `signatureFormat: 'bip137'`, identical r/s, both verifying against that address via `bitcoinMessage.verify(..., checkSegwitAlways: true)`. That confirms the default no longer emits a SegWit header.

Asana: https://app.asana.com/1/9976422036640/project/1215088146871429/task/1216403654258303

---

> [!NOTE]
> **Medium Risk**
> Changes message-signing output for altcoins (prefix fixes) and opt-in BIP137 headers; default Electrum behavior is preserved but wrong-prefix fixes alter signatures that may have been used in production for non-BTC chains.
> 
> **Overview**
> UTXO **message signing** is extended so ramp/CEX ownership proofs (e.g. BIP137 verifiers) can succeed without clients rewriting signatures.
> 
> **`signMessage`** accepts optional `signatureFormat: 'electrum' | 'bip137'` (default **`electrum`**). With **`bip137`**, `signMessageBase64` sets `bitcoinjs-message` **`segwitType`** from the address derivation path (`bip49` → nested SegWit, `bip84` → native SegWit). Signing now uses each coin’s **`messagePrefix`** instead of Bitcoin’s default, and **14 altcoins** had wrong prefix strings/length bytes corrected.
> 
> **`signMessage`** throws **`AddressNotOwnedError`** when the address is invalid for the chain or not owned by the wallet (replacing generic errors).
> 
> Tests add Bitcoin/Litecoin vectors for both encodings and **`messagePrefix.spec`** guards prefix CompactSize bytes across all registered coins.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10d018f9118b247e9fff1ffac6e94b5ab2edbb1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/common/utxobased/keymanager/keymanager.ts
+++ b/src/common/utxobased/keymanager/keymanager.ts
@@ -18,7 +18,13 @@ import { EdgeLog, EdgeMemo } from 'edge-core-js/types'
 
 import { indexAtProtected } from '../../../util/indexAtProtected'
 import { undefinedIfEmptyString } from '../../../util/undefinedIfEmptyString'
-import { ChangePath, CoinInfo, CoinPrefixes, FeeInfo } from '../../plugin/types'
+import {
+  ChangePath,
+  CoinInfo,
+  CoinPrefixes,
+  CurrencyFormat,
+  FeeInfo
+} from '../../plugin/types'
 import { UtxoData } from '../db/types'
 import { ScriptTemplate, ScriptTemplates } from '../info/scriptTemplates/types'
 import { sortInputs, sortOutputs } from './bip69'
@@ -28,7 +34,7 @@ import {
   hashToCashAddress
 } from './bitcoincashUtils/cashAddress'
 import { getCoinFromString } from './coinmapper'
-import { InsufficientFundsErrorPlus } from './types'
+import { InsufficientFundsErrorPlus, UtxoSignatureFormat } from './types'
 import * as utxopicker from './utxopicker'
 import * as pickerUtils from './utxopicker/utils'
 
@@ -938,14 +944,62 @@ export function privateKeyEncodingToPubkey(
   }).publicKey.toString('hex')
 }
 
-export function signMessageBase64(message: string, privateKey: string): string {
+/**
+ * The `segwitType` to hand `bitcoinjs-message`, or `undefined` to leave the
+ * header byte in the legacy Electrum range. Only the BIP137 encoding cares
+ * about the derivation purpose, so only it resolves one.
+ */
+function getBip137SegwitType(
+  signatureFormat: UtxoSignatureFormat,
+  format: CurrencyFormat
+): 'p2wpkh' | 'p2sh(p2wpkh)' | undefined {
+  if (signatureFormat !== 'bip137') return undefined
+  const purposeType = bip43PurposeNumberToTypeEnum(
+    parseInt(format.replace('bip', ''))
+  )
+  switch (purposeType) {
+    case BIP43PurposeTypeEnum.Segwit:
+      return 'p2wpkh'
+    case BIP43PurposeTypeEnum.WrappedSegwit:
+      return 'p2sh(p2wpkh)'
+    default:
+      return undefined
+  }
+}
+
+export function signMessageBase64(
+  message: string,
+  privateKey: string,
+  format: CurrencyFormat,
+  coin: string,
+  signatureFormat: UtxoSignatureFormat = 'electrum'
+): string {
   const ECPair = getECPair()
   const keyPair = ECPair.fromPrivateKey(Buffer.from(privateKey, 'hex'))
   if (keyPair.privateKey == null) {
     throw new Error('Address could not sign message')
   }
+  // BIP137 encodes the address type in the signature's header byte. Native
+  // SegWit (bip84 / P2WPKH) and nested SegWit (bip49 / P2SH-P2WPKH) each have
+  // their own header range; verifiers such as Bringin reject a SegWit address
+  // whose message was signed with a legacy (bip44) header byte. The derivation
+  // path is what tells us the script type, so no caller has to infer it from
+  // the address string. Callers opt in, since emitting a BIP137 header to a
+  // verifier that expects the legacy Electrum encoding is just as broken.
+  // The purpose lookup stays inside the BIP137 branch on purpose:
+  // `bip43PurposeNumberToTypeEnum` throws on a purpose it does not map, so
+  // computing it unconditionally would let a future `CurrencyFormat` (bip86
+  // Taproot being the obvious candidate) break the default Electrum path,
+  // which never needs it.
+  const segwitType = getBip137SegwitType(signatureFormat, format)
+  // Magic-hash the message with the coin's own prefix (e.g. "Litecoin Signed
+  // Message:\n") so the signature verifies against that coin's addresses.
+  // Every UTXO coin routes through here, not just Bitcoin.
+  const messagePrefix = getCoinFromString(coin).prefixes.messagePrefix[0]
   return bitcoinMessage
-    .sign(message, keyPair.privateKey, keyPair.compressed)
+    .sign(message, keyPair.privateKey, keyPair.compressed, messagePrefix, {
+      segwitType
+    })
     .toString('base64')
 }
 
--- a/test/common/utxobased/keymanager/coins/keymanagertest.spec.ts
+++ b/test/common/utxobased/keymanager/coins/keymanagertest.spec.ts
@@ -130,12 +130,22 @@ describe('altcoin test fixtures', () => {
     }
     if (f.signMessageTests != null) {
       f.signMessageTests.forEach(j => {
-        it(`${f.name} sign message test`, () => {
+        const formatName = j.signatureFormat ?? 'default'
+        it(`${f.name} sign message test (${j.format}, ${formatName})`, () => {
           const privateKey = ECPair.fromWIF(j.wif).privateKey?.toString('hex')
           if (privateKey == null) {
             throw new Error('private key cannot be null')
           }
-          const signature = signMessageBase64(j.message, privateKey)
+          const signature =
+            j.signatureFormat == null
+              ? signMessageBase64(j.message, privateKey, j.format, f.name)
+              : signMessageBase64(
+                  j.message,
+                  privateKey,
+                  j.format,
+                  f.name,
+                  j.signatureFormat
+                )
           expect(signature).to.eqls(j.signature)
         })
       })
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ## Unreleased
 
+- added: `signatureFormat` option on `signMessage`, accepting `electrum` (the default) or `bip137`. Existing callers keep the legacy Electrum header byte; BIP137 is opt-in.
+- changed: `signMessage` throws a named `AddressNotOwnedError` when the wallet cannot sign for the requested address, so callers no longer match on error message text.
 - fixed: Correct the `messagePrefix` magic strings for 14 coins so signed messages verify against each chain's own node. Affects dash, pivx, qtum, ravencoin, smartcash, uniformfiscalobject, ecash, zcoin, litecoin, dogecoin, digibyte, feathercoin, bitcoingold and bitcoingoldtestnet.
+- fixed: Encode BIP137 message signatures with the correct header byte for SegWit addresses (bip49 nested SegWit and bip84 native SegWit) and with each coin's own message prefix, so signatures verify on BIP137-compliant platforms across all UTXO currencies.
 
 ## 3.11.0 (2026-07-13)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fixed: Correct the `messagePrefix` magic strings for 14 coins so signed messages verify against each chain's own node. Affects dash, pivx, qtum, ravencoin, smartcash, uniformfiscalobject, ecash, zcoin, litecoin, dogecoin, digibyte, feathercoin, bitcoingold and bitcoingoldtestnet.
+
 ## 3.11.0 (2026-07-13)
 
 - added: Support the `<code>-wif:` protohandler prefix (e.g. `bch-wif:`) in `parseUri` so CashStamps private keys can be swept.
--- a/src/common/utxobased/engine/UtxoEngine.ts
+++ b/src/common/utxobased/engine/UtxoEngine.ts
@@ -45,7 +45,10 @@ import {
   scriptPubkeyToAddress,
   signTx
 } from '../keymanager/keymanager'
-import { asMaybeInsufficientFundsErrorPlus } from '../keymanager/types'
+import {
+  AddressNotOwnedError,
+  asMaybeInsufficientFundsErrorPlus
+} from '../keymanager/types'
 import { transactionSizeFromHex } from '../keymanager/utxopicker/utils'
 import { createPayment, getPaymentDetails, sendPayment } from './paymentRequest'
 import {
@@ -797,11 +800,20 @@ export async function makeUtxoEngine(
       opts: EdgeSignMessageOptions
     ): Promise<string> {
       const otherParams = asUtxoSignMessageOtherParams(opts.otherParams)
-      const { publicAddress } = otherParams
-      const scriptPubkey = walletTools.addressToScriptPubkey(publicAddress)
+      const { publicAddress, signatureFormat } = otherParams
+      // An address the wallet cannot sign for reaches us two ways: it parses
+      // but is not ours, or it is not an address of this chain at all and
+      // `addressToScriptPubkey` throws. Both are the same thing to a caller,
+      // so give them one named error rather than assorted parser prose.
+      let scriptPubkey: string
+      try {
+        scriptPubkey = walletTools.addressToScriptPubkey(publicAddress)
+      } catch {
+        throw new AddressNotOwnedError()
+      }
       const addressData = await dataLayer.fetchAddress(scriptPubkey)
       if (addressData?.path == null) {
-        throw new Error('Missing data-layer address to sign with')
+        throw new AddressNotOwnedError()
       }
       const privateKey = asMaybeCurrencyPrivateKey(privateKeys)
 
@@ -818,7 +830,8 @@ export async function makeUtxoEngine(
       const signature = await walletTools.signMessageBase64({
         path: addressData?.path,
         message,
-        xprivKeys
+        xprivKeys,
+        signatureFormat
       })
       return signature
     },
--- a/src/common/utxobased/engine/UtxoWalletTools.ts
+++ b/src/common/utxobased/engine/UtxoWalletTools.ts
@@ -19,6 +19,7 @@ import {
   xprivToPrivateKey,
   xpubToPubkey
 } from '../keymanager/keymanager'
+import { UtxoSignatureFormat } from '../keymanager/types'
 import {
   CurrencyFormatKeys,
   currencyFormatToPurposeType,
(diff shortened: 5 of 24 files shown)


j0ntz:
## 📸 Test evidence: in-app BIP137 signatures (iOS sim, edge-funds BTC wallet)

**agent proof 1216403654258303 01 bip137 signatures**

agent proof 1216403654258303 01 bip137 signatures

_Captured by the agent's in-app test run (build-and-test)._

j0ntz (commented):
Automated code review (workflow-backed, high effort). All findings are advisory; the core BIP137 header fix is correct and matches the task requirement. Three notes below.

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -938,14 +944,28 @@ export function privateKeyEncodingToPubkey(
...
+  // BIP137 encodes the address type in the signature's header byte. Native
+  // SegWit (bip84 / P2WPKH) and nested SegWit (bip49 / P2SH-P2WPKH) each have
+  // their own header range; verifiers such as Bringin reject a SegWit address
+  // whose message was signed with a legacy (bip44) header byte.
+  const segwitType =
+    format === 'bip84'
+      ? 'p2wpkh'
+      : format === 'bip49'
+      ? 'p2sh(p2wpkh)'
+      : undefined
```
j0ntz: **Scope: this path is shared by every UTXO coin.** `signMessageBase64` is reached from `UtxoWalletTools.signMessage` for all UTXO currencies, so bip49/bip84 wallets of altcoins (Litecoin, Digibyte, etc.) now also emit BIP137 SegWit header bytes, not just Bitcoin. Two consequences worth being aware of:

1. **Altcoin magic prefix is still Bitcoin's.** The digest is magic-hashed with bitcoinjs-message's default `"Bitcoin Signed Message:"` prefix regardless of coin, so a non-Bitcoin SegWit signature now has the correct header but the wrong (Bitcoin) prefix. This prefix behavior is pre-existing (not introduced here) and the task is BTC-scoped, so no action is required, but a coin whose partner starts requiring BIP137 will need the coin-specific prefix threaded through too.
2. **Legacy-only verifiers.** BTC bip49/bip84 signatures now carry SegWit headers (`I.../J...`), which Bitcoin Core `verifymessage` and Electrum's default verify do not accept. This is the intended BIP137 behavior (Bringin requires it) and the correct tradeoff, just flagging that any internal/legacy verifier of Edge-produced signatures must be BIP137-aware.

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -938,14 +944,28 @@ export function privateKeyEncodingToPubkey(
...
   const ECPair = getECPair()
   const keyPair = ECPair.fromPrivateKey(Buffer.from(privateKey, 'hex'))
   if (keyPair.privateKey == null) {
     throw new Error('Address could not sign message')
   }
+  // BIP137 encodes the address type in the signature's header byte. Native
+  // SegWit (bip84 / P2WPKH) and nested SegWit (bip49 / P2SH-P2WPKH) each have
+  // their own header range; verifiers such as Bringin reject a SegWit address
+  // whose message was signed with a legacy (bip44) header byte.
+  const segwitType =
```
j0ntz: **Minor (robustness):** the `format` -> `segwitType` mapping compares raw `'bip84'`/`'bip49'` string literals rather than routing through the codebase's canonical `BIP43PurposeTypeEnum` / `bip43PurposeNumberToTypeEnum`. If a new SegWit `CurrencyFormat` is ever added, this ternary is easy to miss (no compiler linkage) and the new format silently falls through to `undefined`, producing a legacy header that BIP137 verifiers reject. Consider deriving the segwit kind from the purpose enum so the two stay in lockstep.

test/common/utxobased/keymanager/coins/keymanagertest.spec.ts
```diff
@@ -130,12 +130,12 @@ describe('altcoin test fixtures', () => {
...
     if (f.signMessageTests != null) {
       f.signMessageTests.forEach(j => {
-        it(`${f.name} sign message test`, () => {
+        it(`${f.name} sign message test (${j.format})`, () => {
           const privateKey = ECPair.fromWIF(j.wif).privateKey?.toString('hex')
           if (privateKey == null) {
             throw new Error('private key cannot be null')
           }
-          const signature = signMessageBase64(j.message, privateKey)
+          const signature = signMessageBase64(j.message, privateKey, j.format)
```
j0ntz: **Test strength:** the bip49/bip84 expected signatures in `altcointestfixtures.ts` are the bip44 fixture with only the leading base64 char hand-incremented (`H` -> `I` -> `J`), and this assertion is pure string equality. That proves the header byte shifted, but not that the signature actually verifies for a SegWit address. Consider adding a `bitcoinMessage.verify(message, address, signature)` round-trip against the real P2SH-P2WPKH / bech32 address so a subtly-wrong recovery flag can't be masked by editing the fixture to match. (This run verified the round-trip offline; folding it into the suite would lock it in.)

src/common/utxobased/keymanager/keymanager.ts
j0ntz: Addressed. `signMessageBase64` now threads each coin's own `messagePrefix` (`getCoinFromString(coin).prefixes.messagePrefix[0]`) into `bitcoinMessage.sign`, so every UTXO coin magic-hashes with its own prefix (e.g. "Litecoin Signed Message:") instead of always Bitcoin's. This is the task's actual scope: all UTXO currencies, not only BTC. Added litecoin bip44/49/84 sign-message fixtures whose signatures were independently validated with `bitcoinMessage.verify` against the derived litecoin legacy/p2sh/bech32 addresses before hardcoding.

test/common/utxobased/keymanager/coins/keymanagertest.spec.ts
j0ntz: Partly addressed. The suite now covers litecoin bip44/49/84 in addition to bitcoin, and each new signature was validated with a `bitcoinMessage.verify(message, address, signature, prefix, true)` round-trip against the coin-derived address during vector generation. The in-suite assertion is still string equality; folding the verify() call directly into the spec remains a reasonable future hardening.

src/common/utxobased/keymanager/keymanager.ts
j0ntz: Leaving open as an optional cleanup. The coin-prefix fix (the operator's requested scope) does not depend on it, so this change stays minimal; routing the segwitType selection through `BIP43PurposeTypeEnum` is a good follow-up but out of scope for this task.

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -938,14 +944,35 @@ export function privateKeyEncodingToPubkey(
...
+    format === 'bip84'
+      ? 'p2wpkh'
+      : format === 'bip49'
+      ? 'p2sh(p2wpkh)'
+      : undefined
+  // Magic-hash the message with the coin's own prefix (e.g. "Litecoin Signed
+  // Message:\n") so the signature verifies against that coin's addresses.
+  // Every UTXO coin routes through here, not just Bitcoin.
+  const messagePrefix = getCoinFromString(coin).prefixes.messagePrefix[0]
```
peachbits, line 998: The messagePrefixes are mostly wrong. Here's a fix for them https://github.com/EdgeApp/edge-currency-plugins/commit/6591ad80d9213c19e95d519b20b83b3d750cc85c

The LTC tests are wrong and need to be redone on top of that

peachbits (changes requested):

src/common/utxobased/keymanager/keymanager.ts
j0ntz: Done, both parts.

Absorbed your `6591ad80` into this PR with authorship preserved, ordered BEFORE the BIP137 commit so each commit is self-consistent: the prefix data is corrected first, then the signer starts consuming it. (No PR existed on `matthew/audit-message-prefixes`, so I pulled the commit in here rather than waiting on a separate land. Say the word if you would rather land it standalone and have me rebase.)

Your point about the LTC vectors was exactly right, and the pre-commit suite caught it the moment the prefix flipped: `\x18` -> `\x19` invalidated all three. Regenerated them against `litecoin.coinInfo.prefixes.messagePrefix[0]` read straight from the coin info, and validated each with `bitcoinMessage.verify` against the derived addresses (`LZGpRyQPybaDjbRGoB87YH2ebFnmKYmRui`, `MKzeSBgnLcZfN96jgm8BacGPviggPBSJ58`, `ltc1qngw83fg8dz0k749cg7k3emc7v98wy0c73398wa`) rather than pasting the failure output back in. Also asserted the negative: none of the three verify under Bitcoin's prefix, so the vectors would catch a regression that silently reverted to the default.

Your `messagePrefix.spec.ts` guard now runs over 24 coins and passes; full verify-repo green at 1245 tests.

src/common/utxobased/keymanager/keymanager.ts
j0ntz: Implemented after all. The prefix audit in this PR was the same failure shape this thread describes (inert data that silently produces unverifiable signatures), so the linkage is worth having: `segwitType` now derives from `bip43PurposeNumberToTypeEnum(...)` compared against `BIP43PurposeTypeEnum.Segwit` / `.WrappedSegwit` instead of raw `'bip84'`/`'bip49'` literals. A newly-added format now fails loudly in the purpose mapping rather than falling through to a legacy header.

Used the local enum helper rather than `currencyFormatToPurposeType` from `engine/utils`, since that module already imports from keymanager and the reverse import would close a cycle. Behavior is identical across bip32/44/49/84 and the full suite is green.

j0ntz:
Updated in 4cb6e2bb to resolve the cross-PR conflict raised on [EdgeApp/edge-react-gui#6065](https://github.com/EdgeApp/edge-react-gui/pull/6065).

BIP-137 is now **opt-in** rather than unconditional. `asUtxoSignMessageOtherParams` carries `signatureFormat: 'electrum' | 'bip137'`, defaulting to `electrum`, threaded through `UtxoEngine.signMessage` → `UtxoWalletTools.signMessageBase64` → `keymanager.signMessageBase64`, which applies `segwitType` only when BIP-137 is requested. Existing callers (`bityProvider`, `EdgeProviderServer`) keep the header byte they produce today, so this stops being a silent behavior change for them and merge order with #6065 no longer matters.

Also here:
- `signMessage` throws a named `AddressNotOwnedError` for an address the wallet cannot sign for, whether it fails to parse or simply is not ours. Callers branch on `error.name` instead of matching parser prose (same pattern as `asMaybeInsufficientFundsErrorPlus`).
- The keymanager spec covers both directions on bitcoin and litecoin: the default encoding yields the same `H…` header for bip44/bip49/bip84, and only an explicit `bip137` request shifts bip49 to `I…` and bip84 to `J…`. 12 cases, so swapping the two formats fails the suite.

`npm run verify` passes locally (1251 tests). Driven end to end on the iOS sim with this build linked into the GUI: signing with a bip84 `bc1q…` address produced header 32 under the default and header 40 under BIP-137, identical r/s, both verifying against that address via `bitcoinMessage.verify(..., checkSegwitAlways: true)`.

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -938,14 +944,44 @@ export function privateKeyEncodingToPubkey(
...
+  // BIP137 encodes the address type in the signature's header byte. Native
+  // SegWit (bip84 / P2WPKH) and nested SegWit (bip49 / P2SH-P2WPKH) each have
+  // their own header range; verifiers such as Bringin reject a SegWit address
+  // whose message was signed with a legacy (bip44) header byte. The derivation
+  // path is what tells us the script type, so no caller has to infer it from
+  // the address string. Callers opt in, since emitting a BIP137 header to a
+  // verifier that expects the legacy Electrum encoding is just as broken.
+  const purposeType = bip43PurposeNumberToTypeEnum(
+    parseInt(format.replace('bip', ''))
+  )
```
peachbits: `purposeType` is computed before the `signatureFormat !== 'bip137'` check, but only the BIP137 branch reads it. Since `bip43PurposeNumberToTypeEnum` throws on an unmapped purpose, adding a fifth `CurrencyFormat` — bip86 Taproot being the obvious candidate — would break **all** message signing, including the default Electrum path that never touches `purposeType`.

Folding it into the bip137 branch keeps the failure scoped to the format that actually depends on it. Not reachable today, since `asCurrencyFormat` is a closed union of exactly bip32/44/49/84 — this is about the next format, not this PR.

j0ntz: Fixed. The purpose lookup moved into a `getBip137SegwitType` helper that returns early on any non-`bip137` format, so `bip43PurposeNumberToTypeEnum` is only reached when the BIP137 encoding actually needs it. A future `CurrencyFormat` it does not map now breaks only BIP137 signing for that format, not the default Electrum path.

The switch also replaced the nested ternary, so an unmapped-but-known purpose falls through to `undefined` (legacy header) rather than being buried in the chain.

Existing coverage holds it: the 12 keymanager cases still assert the default encoding produces the same `H…` header for bip44/bip49/bip84, so a regression that made the default path depend on purpose resolution would fail there.

src/common/utxobased/keymanager/keymanager.ts
```diff
@@ -938,14 +944,44 @@ export function privateKeyEncodingToPubkey(
...
+  const segwitType =
+    signatureFormat !== 'bip137'
+      ? undefined
+      : purposeType === BIP43PurposeTypeEnum.Segwit
+      ? 'p2wpkh'
+      : purposeType === BIP43PurposeTypeEnum.WrappedSegwit
+      ? 'p2sh(p2wpkh)'
+      : undefined
```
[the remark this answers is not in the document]
j0ntz: Valid, and already fixed in 2bc9171 (same finding peachbits raised). The purpose lookup now lives in a `getBip137SegwitType` helper that returns `undefined` before touching `bip43PurposeNumberToTypeEnum` whenever `signatureFormat !== 'bip137'`, so an unmapped future `CurrencyFormat` can only break BIP137 signing, never the default Electrum path.

This review ran against d8f9bbc, which is the pre-fix HEAD that the fixup squash produced.

peachbits (approved):

PR #455, opened 2026-08-04
Never report a successful broadcast as a failed send

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

[Technical design doc](https://github.com/EdgeApp/edge-react-gui/blob/jon/send-post-broadcast-failure/src/docs/send-post-broadcast-failure.md)

Fixes the engine half of Edge bug "Send - multiple transactions after network error": one intended Bitcoin send became five real payments because every failure the user saw happened AFTER a successful broadcast.

Asana: https://app.asana.com/1/9976422036640/project/1213843652804305/task/1217135300337949

Two defects fixed:

1. `saveTx` failed on a disconnected engine. `updateProgressRatio` threw `No addresses to process` whenever zero addresses were subscribed (exactly the state of a wallet whose blockbook sockets are down), and `saveTx` reaches it via `processUtxos` after the transaction is already saved and its inputs marked spent. The throw is progress bookkeeping with no denominator, not a data error, so it is now a no-op instead of failing the caller's data write. Regression test: `saveTx` on a never-connected engine resolves (`test/common/utxobased/engine/saveTx.spec.ts`, red before the fix with the exact incident stack).

2. Broadcast failure was ambiguous. `ServerStates.broadcastTx` multicasts to every connected blockbook (or every NOWNode HTTP fallback) and rejects only when all fail, but a server can relay the transaction and still return an error or time out, so "all servers errored" does not prove the transaction is absent from the network. The exhausted broadcast is now CLASSIFIED, with no added network calls: an already-known rejection (already in block chain / txn-already-in-mempool / txn-already-known) resolves the broadcast as a SUCCESS, since the server's own answer proves the transaction reached the network; when every relay-capable failure is an explicit Blockbook rejection the engine rejects with the original error (definitively failed, safe to retry); when any failure is a transport error it rejects with `BroadcastAmbiguityError`, which the GUI will use to lock the retry path and show may-have-worked messaging (tracked separately). Electrum stub refusals, which provably never sent anything, are excluded from the determination. An earlier revision queried the network for the txid before rejecting; that was removed after review, since an immediate query runs under the same network conditions that made the broadcast ambiguous and proves nothing either way. The post-broadcast txid-mismatch throw in `UtxoEngine.broadcastTx` (dead code today, failure-after-success if ever revived) now logs a warning and returns the transaction with the network's txid so the wallet tracks what was actually accepted.

The GUI half (send scene must not present post-broadcast errors as failed sends) is the EdgeApp/edge-react-gui companion PR on branch `jon/send-post-broadcast-failure`.

---

> [!NOTE]
> **High Risk**
> Changes UTXO send/broadcast and saveTx behavior on the payment path; incorrect classification could still allow duplicate sends or block legitimate retries.
> 
> **Overview**
> Fixes cases where a **successful broadcast** was still reported as a failed send, which led to duplicate payments when users retried.
> 
> **`saveTx` on a disconnected engine** no longer fails after the transaction is already saved. When no addresses are subscribed, `updateProgressRatio` skips progress updates instead of throwing `No addresses to process`. **`stop()`** resets progress counters so leftover counts cannot emit a bogus fully-synced state during later `saveTx`-driven UTXO work.
> 
> **Exhausted multicasts** in `ServerStates.broadcastTx` are classified via new `broadcastError` helpers: all explicit Blockbook rejections still reject with the original error (safe to retry); any transport-style failure rejects with **`BroadcastAmbiguityError`** so the UI can treat the send as possibly on-network. **Already-known** mempool/chain rejections are treated as success. **`UtxoEngine.broadcastTx`** logs and returns the network’s txid on mismatch instead of throwing after relay.
> 
> Tests cover classification (`broadcastError.spec.ts`) and disconnected-engine `saveTx` (`saveTx.spec.ts`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51e5b66fd16234ffdb34709b8ea4095362b30700. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/common/utxobased/engine/ServerStates.ts
+++ b/src/common/utxobased/engine/ServerStates.ts
@@ -1,3 +1,4 @@
+import { asMaybe, asObject, asString } from 'cleaners'
 import { EdgeIo, EdgeLog, EdgeTransaction } from 'edge-core-js/types'
 import { parse } from 'uri-js'
 
@@ -367,9 +368,75 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
 
   const instance: ServerStates = {
     async broadcastTx(transaction: EdgeTransaction): Promise<string> {
+      // Query the network for the transaction to determine whether a failed
+      // broadcast actually reached the network anyway. A server can relay
+      // the transaction and still return an error or fail to respond, so an
+      // error from every server does not prove the transaction wasn't sent.
+      const isTxidKnown = async (txid: string): Promise<boolean> => {
+        // Ask connected blockbook instances first:
+        for (const uri of Object.keys(serverStatesCache)) {
+          const { blockbook } = serverStatesCache[uri]
+          if (blockbook == null || !blockbook.isConnected) continue
+          const known = await blockbook
+            .fetchTransaction(txid)
+            .then(() => true)
+            .catch(() => false)
+          if (known) return true
+        }
+
+        // Fall back to the NOWNode HTTP API when no blockbook is connected:
+        const { nowNodesApiKey } = initOptions
+        if (nowNodesApiKey == null) return false
+        const nowNodeUris = serverConfigs
+          .filter(config => config.type === 'blockbook-nownode')
+          .map(config => config.uris)
+          .flat(1)
+        for (const uri of nowNodeUris) {
+          const known = await io
+            .fetchCors(`${uri}/api/v2/tx/${txid}`, {
+              headers: {
+                'api-key': nowNodesApiKey
+              }
+            })
+            .then(async response => {
+              if (!response.ok) return false
+              const json = await response.json()
+              return asMaybe(asTxQueryResponse)(json)?.txid === txid
+            })
+            .catch(() => false)
+          if (known) return true
+        }
+        return false
+      }
+
       return await new Promise((resolve, reject) => {
         let resolved = false
         let bad = 0
+
+        // Reject with the given error only when the transaction is verifiably
+        // absent from the network; a transaction that reached the network
+        // despite the error is a successful broadcast.
+        const rejectUnlessTxKnown = (error?: Error): void => {
+          const fail = (): void => {
+            const msg = error != null ? `With error ${error.message}` : ''
+            log.error(
+              `broadcastTx fail: ${JSON.stringify(transaction)}\n${msg}`
+            )
+            reject(error)
+          }
+          isTxidKnown(transaction.txid)
+            .then(known => {
+              if (!known) return fail()
+              if (!resolved) {
+                resolved = true
+                log.warn(
+                  `broadcastTx errored, but txid ${transaction.txid} is known to the network; treating broadcast as a success`
+                )
+                resolve(transaction.txid)
+              }
+            })
+            .catch(fail)
+        }
 
         const wsUris = Object.keys(serverStatesCache).filter(
           uri => serverStatesCache[uri].blockbook != null
@@ -394,11 +461,7 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
               })
               .catch((e?: Error) => {
                 if (++bad === wsUris.length) {
-                  const msg = e != null ? `With error ${e.message}` : ''
-                  log.error(
-                    `broadcastTx fail: ${JSON.stringify(transaction)}\n${msg}`
-                  )
-                  reject(e)
+                  rejectUnlessTxKnown(e)
                 }
               })
           }
@@ -464,11 +527,7 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
               })
               .catch((e?: Error) => {
                 if (++bad === nowNodeUris.length) {
-                  const msg = e != null ? `With error ${e.message}` : ''
-                  log.error(
-                    `broadcastTx fail: ${JSON.stringify(transaction)}\n${msg}`
-                  )
-                  reject(e)
+                  rejectUnlessTxKnown(e)
                 }
               })
           }
@@ -670,3 +729,11 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
 
   return instance
 }
+
+/**
+ * Minimal shape of a Blockbook REST `/api/v2/tx/<txid>` response, used only
+ * to confirm that a transaction is known to the network.
+ */
+const asTxQueryResponse = asObject({
+  txid: asString
+})
--- a/src/common/utxobased/engine/ServerStates.ts
+++ b/src/common/utxobased/engine/ServerStates.ts
@@ -22,7 +22,8 @@ import { SocketEmitter, SocketEvent } from '../network/SocketEmitter'
 import { pushUpdate, removeIdFromQueue } from '../network/socketQueue'
 import {
   BroadcastAmbiguityError,
-  classifyBroadcastFailure
+  classifyBroadcastFailure,
+  isAlreadyKnownRejection
 } from './broadcastError'
 import { MAX_CONNECTIONS, NEW_CONNECTIONS } from './constants'
 import { UtxoInitOptions } from './types'
@@ -382,6 +383,19 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
         // have relayed the transaction before failing to answer.
         const broadcastErrors: unknown[] = []
         const rejectClassified = (error?: Error): void => {
+          // A server refusing because it already has the transaction is
+          // confirmation the transaction reached the network (from this
+          // attempt or an earlier one with the same signed bytes): success.
+          if (broadcastErrors.some(isAlreadyKnownRejection)) {
+            if (!resolved) {
+              resolved = true
+              log.warn(
+                `broadcastTx: server already has txid ${transaction.txid}; treating broadcast as a success`
+              )
+              resolve(transaction.txid)
+            }
+            return
+          }
           const msg = error != null ? `With error ${error.message}` : ''
           log.error(`broadcastTx fail: ${JSON.stringify(transaction)}\n${msg}`)
           if (classifyBroadcastFailure(broadcastErrors) === 'ambiguous') {
--- a/src/common/utxobased/engine/ServerStates.ts
+++ b/src/common/utxobased/engine/ServerStates.ts
@@ -1,4 +1,3 @@
-import { asMaybe, asObject, asString } from 'cleaners'
 import { EdgeIo, EdgeLog, EdgeTransaction } from 'edge-core-js/types'
 import { parse } from 'uri-js'
 
@@ -21,6 +20,10 @@ import Deferred from '../network/Deferred'
 import { WsTask, WsTaskGenerator } from '../network/Socket'
 import { SocketEmitter, SocketEvent } from '../network/SocketEmitter'
 import { pushUpdate, removeIdFromQueue } from '../network/socketQueue'
+import {
+  BroadcastAmbiguityError,
+  classifyBroadcastFailure
+} from './broadcastError'
 import { MAX_CONNECTIONS, NEW_CONNECTIONS } from './constants'
 import { UtxoInitOptions } from './types'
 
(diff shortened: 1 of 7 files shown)


src/common/utxobased/engine/ServerStates.ts
```diff
@@ -367,10 +368,76 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
...
+            .then(() => true)
+            .catch(() => false)
+          if (known) return true
+        }
+
+        // Fall back to the NOWNode HTTP API when no blockbook is connected:
+        const { nowNodesApiKey } = initOptions
+        if (nowNodesApiKey == null) return false
+        const nowNodeUris = serverConfigs
+          .filter(config => config.type === 'blockbook-nownode')
```
j0ntz: Why this query is hard-coded to NOWNode: the checker must work in the exact state the incident happened in (zero connected blockbooks), and the only transport the engine is configured to reach there is the NOWNode REST API. The sendtx fallback ~90 lines below already restricts itself to `type === 'blockbook-nownode'` + `nowNodesApiKey` (with a commented-out future hook for user-configured HTTP servers), so the query mirrors that exactly: whatever server could have relayed the transaction is the server we ask about it. Connected blockbooks are tried first, so NOWNode is only the last resort for the disconnected state.

Failure direction is safe: with no NOWNode config or key, `isTxidKnown` returns false and the broadcast rejects with the original error, which is the pre-fix behavior. The hard-coding can cost a missed rescue, never a false success.

Known drift risk: the sendtx fallback and this query each derive the NOWNode uri list + key independently. If the fallback ever broadens to other HTTP servers, this query must broaden with it; a shared `resolveHttpFallbackServers()` helper would collapse the two sites and is a clean follow-up.

src/common/utxobased/engine/ServerStates.ts
```diff
@@ -367,10 +368,76 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
...
+        // Ask connected blockbook instances first:
+        for (const uri of Object.keys(serverStatesCache)) {
+          const { blockbook } = serverStatesCache[uri]
+          if (blockbook == null || !blockbook.isConnected) continue
+          const known = await blockbook
+            .fetchTransaction(txid)
+            .then(() => true)
+            .catch(() => false)
+          if (known) return true
+        }
```
peachbits: **Unbounded latency on the send path.** This loop `await`s each blockbook serially, and `Socket.ts` uses a 30s per-request timeout (`constants.ts` `MAX_CONNECTIONS = 2`), so this alone can add ~60s. The NOWNodes loop below then adds N more `io.fetchCors` calls with *no* timeout at all.

Before this change a total broadcast failure rejected immediately. Now the user sits on a spinner for a minute or more — potentially forever if a `fetchCors` never settles — before being told the send failed.

Suggest racing these concurrently and wrapping the whole `isTxidKnown` call in an overall deadline (~5-10s), falling back to `fail()` on timeout.

src/common/utxobased/engine/ServerStates.ts
```diff
@@ -367,10 +368,76 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
...
+
+        // Fall back to the NOWNode HTTP API when no blockbook is connected:
+        const { nowNodesApiKey } = initOptions
+        if (nowNodesApiKey == null) return false
+        const nowNodeUris = serverConfigs
+          .filter(config => config.type === 'blockbook-nownode')
+          .map(config => config.uris)
+          .flat(1)
```
peachbits: The comment says *"when no blockbook is connected"*, but there is no such condition — this block runs on **every** exhausted broadcast, including the case where blockbooks are connected and simply returned `false` above.

Either gate it on `!isAnyBlockbookConnected` to match the intent (and match the broadcast fallback at L472), or fix the comment. As-is it silently widens both the latency and the api-key exposure below.

src/common/utxobased/engine/ServerStates.ts
```diff
@@ -367,10 +368,76 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
...
+              }
+            })
+            .then(async response => {
+              if (!response.ok) return false
+              const json = await response.json()
+              return asMaybe(asTxQueryResponse)(json)?.txid === txid
+            })
+            .catch(() => false)
+          if (known) return true
+        }
```
peachbits: **The NOWNodes api-key goes to hosts that aren't NOWNodes.** `ServerConfig['type']` is a single-member union (`types.ts:160`), so `.filter(config => config.type === 'blockbook-nownode')` never excludes anything. And those URI lists are mixed — `bitcoin.ts:54` has `https://btc-wusa1.edge.app`, `https://btc-eu1.edge.app` *and* `https://btcbook.nownodes.io`, all under the same type. Same in `bitcoincash.ts`, etc.

This is pre-existing in the broadcast fallback at L502-512, so not introduced here — but this PR replicates it on a new path that (per the comment above) fires far more often. Worth splitting the config type so the key is only attached to actual `nownodes.io` hosts, rather than duplicating the pattern.

src/common/utxobased/engine/ServerStates.ts
```diff
@@ -367,10 +368,76 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
...
+              if (!known) return fail()
+              if (!resolved) {
+                resolved = true
+                log.warn(
+                  `broadcastTx errored, but txid ${transaction.txid} is known to the network; treating broadcast as a success`
+                )
+                resolve(transaction.txid)
+              }
+            })
+            .catch(fail)
```
peachbits: **No propagation delay means this will usually return `false` when it matters.** `isTxidKnown` fires the instant the last broadcast rejects. In the exact race being defended against — the node relayed the tx but the response was lost — the tx has had ~0ms to be indexed into the queried server's mempool, so `fetchTransaction` / `/api/v2/tx/` will very likely 404 and we `fail()` anyway.

A short delay plus one retry before giving up would make the check actually fire in the target scenario. Without it, the added latency buys little.

Minor: an early `if (resolved) return` at the top would also skip the whole query in the (currently unreachable, but cheap to guard) case where another server already resolved.

src/common/utxobased/engine/UtxoEngineProcessor.ts
```diff
@@ -170,8 +170,12 @@ export function makeUtxoEngineProcessor(
...
     processedCount = processedCount + 1
 
-    // If we have no addresses, we should not have not yet began processing.
-    if (expectedProcessCount === 0) throw new Error('No addresses to process')
+    // With no subscribed addresses there is no denominator to compute a
+    // progress ratio from. This is a legitimate state when processing is
+    // driven by saveTx on a disconnected engine (no blockbook sockets, so
+    // nothing is subscribed), so skip the progress update rather than fail
+    // the caller's data write.
+    if (expectedProcessCount === 0) return
```
peachbits, line 176: Returning instead of throwing is right, but two things here:

**1. The comment's premise is wrong.** This is *not* "a running engine whose sockets are all down." `initializeAddressSubscriptions()` (L300) populates `addressSubscribeCache` from the DataLayer with no network involvement, and entries are only ever marked `processing`, never removed (L936-966). The cache is empty in exactly two states: the engine was **never started**, or it was **stopped** (`clearTaskCache` at L136). A started-but-disconnected engine has a full cache. Worth correcting here and in `saveTx.spec.ts:112-113`, since it changes how someone reasons about this branch.

**2. `setLookAhead` now runs in a state it never reached before, and the counter is already dirty.** Previously the throw on L177 short-circuited `processDataLayerUtxos` before `setLookAhead(common)` at L1583. Now it runs. `setLookAhead` itself is network-free and safe, but it repopulates `addressSubscribeCache` with the handful of addresses it newly derives — so the denominator goes from `0` to something very small.

Combined with `processedCount` being incremented *above* the guard on L171, a single `saveTx` with two scriptPubkeys (input + change — the common case) does:
- call 1: `expectedProcessCount = 0`, `processedCount → 1`, return; `setLookAhead` derives 1 new address → cache size 1
- call 2: `expectedProcessCount = 2`, `processedCount → 2` → `percent === 1`

That emits `ADDRESSES_CHECKED(1)` and calls `updateSeenTxCheckpoint()` on a stopped engine, advancing the seen-tx checkpoint to `maxSeenTxBlockHeight` without having synced. Moving the increment below the guard fixes it — a call with no denominator shouldn't count as progress.

src/common/utxobased/engine/UtxoEngine.ts
```diff
@@ -432,7 +432,11 @@ export async function makeUtxoEngine(
         throw err
       })
       if (id !== transaction.txid) {
-        throw new Error('broadcast response txid does not match original')
+        // The transaction is on the network at this point, so a mismatched
+        // response txid must not be reported as a send failure.
+        log.warn(
+          `broadcast response txid mismatch: expected ${transaction.txid} received ${id}`
+        )
```
peachbits, line 441: Agreed that throwing here was wrong post-broadcast, but returning `transaction` unchanged means the wallet persists `transaction.txid` while the network accepted `id`. The wallet then tracks a txid that doesn't exist on-chain: it never confirms, the UTXOs stay marked spent, and the funds look stuck — with no error surfaced anywhere the user can see.

Since `formats` includes non-segwit (`bip44`, `bip32`), a genuine txid change isn't purely hypothetical. Consider returning `{ ...transaction, txid: id }` so the wallet tracks what the network actually has, or at minimum promote this above `log.warn`.

test/common/utxobased/engine/saveTx.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+      },
+      log: testLog,
+      infoPayload: {},
+      nativeIo,
+      pluginDisklet: fakeIoDisklet
+    }
+    const factory = edgeCorePlugins[tests.pluginId]
+    if (typeof factory !== 'function')
+      throw new Error(`Missing plugin factory for ${tests.pluginId}`)
+    const plugin: EdgeCurrencyPlugin = factory(pluginOpts) as any
```
peachbits: `const plugin: EdgeCurrencyPlugin = factory(pluginOpts) as any` — the annotation and the `as any` cancel each other out. `factory` returns `EdgeCorePlugin`; narrowing with a type guard (or `as EdgeCurrencyPlugin`) keeps the assertion visible instead of disabling checking entirely.

peachbits (changes requested):
The first commit fixes a real crash, but the `return` lands in a spot that lets `setLookAhead` run in a state it previously never reached — see the `UtxoEngineProcessor.ts` comment. The second commit's verification step is defensible in principle (a node can relay a tx and still error), but as written it is serial, untimed, and unconditional, so it turns a fast send failure into a long one on the money path.

src/common/utxobased/engine/ServerStates.ts
j0ntz: Superseded: the txid query is removed entirely per the review discussion (commit 7b6c8e2 replaces it with failure classification), so the NOWNode coupling this comment defended no longer exists.

src/common/utxobased/engine/ServerStates.ts
j0ntz: Addressed by removing the query (7b6c8e2): the replacement classifies the collected failure set by error shape, adds zero network calls, and the send path rejects exactly as fast as before this PR.

src/common/utxobased/engine/ServerStates.ts
j0ntz: Moot with the query removed (7b6c8e2); no connectivity-conditional behavior remains on the rejection path.

src/common/utxobased/engine/ServerStates.ts
j0ntz: The query path is removed (7b6c8e2), so this PR no longer adds any api-key exposure. The pre-existing exposure in the broadcast fallback is unchanged; splitting ServerConfig so the key only reaches nownodes.io hosts is recorded as a follow-up in the run report.

src/common/utxobased/engine/ServerStates.ts
j0ntz: Agreed, and this is why the query is removed rather than tuned (7b6c8e2): any delay constant would be arbitrary and the engine's ordinary sync already surfaces the transaction when any server sees it. The replacement classifies the failure set with no timing assumption; a set containing any transport failure rejects as BroadcastAmbiguityError so the GUI can lock the retry path (the GUI branch is tracked separately).

src/common/utxobased/engine/UtxoEngineProcessor.ts
j0ntz: Both fixed in 11c8184: the increment moved below the guard so a denominator-less call no longer counts as progress, and the comment here plus the spec's now describe the real empty-cache states (never started, or stopped with the task cache cleared). The spec runs a second saveTx and asserts no fully-synced emission and no seen-tx checkpoint advance; those assertions are invariant guards rather than position-discriminating (the fixture's setLookAhead derives a large lookahead set, so two calls stay far from ratio 1 under either increment order).

src/common/utxobased/engine/UtxoEngine.ts
j0ntz: Fixed in 9d2a7e8: on mismatch the engine now returns { ...transaction, txid: id } so the wallet tracks the txid the network accepted, keeping the warning for diagnostics.

test/common/utxobased/engine/saveTx.spec.ts
j0ntz: Fixed in 11c8184: the annotation-plus-as-any is now a single visible assertion, factory(pluginOpts) as EdgeCurrencyPlugin, behind the existing typeof guard.

j0ntz:
Both review points addressed: the saveTx return no longer counts denominator-less calls as progress and setLookAhead's new reachability is guarded by spec assertions (11c8184); the verification step is removed entirely in favor of zero-latency failure classification (7b6c8e2), so the send path rejects as fast as before, with ambiguous transport failures now distinguishable from explicit rejections.

src/common/utxobased/engine/UtxoEngineProcessor.ts
```diff
@@ -167,12 +167,17 @@ export function makeUtxoEngineProcessor(
     const expectedProcessCount =
       Object.keys(taskCache.addressSubscribeCache).length * processesPerAddress
 
+    // With no subscribed addresses there is no denominator to compute a
+    // progress ratio from. The cache is empty when the engine is not running
+    // (never started, or stopped and its task cache cleared) while saveTx
+    // still drives UTXO processing. Skip the progress update, without
+    // counting the call as progress, rather than fail the caller's data
+    // write.
+    if (expectedProcessCount === 0) return
```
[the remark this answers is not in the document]
j0ntz: Valid, fixed in the updated fixup (branch head b5255c0): stop() now resets processedCount and processedPercent alongside clearing the task cache, so leftover counters from an ended sync round cannot combine with a smaller refilled cache (saveTx-driven setLookAhead or addGapLimitAddresses) into a completed-sync emission or a checkpoint advance.

src/common/utxobased/engine/UtxoEngine.ts
```diff
@@ -432,7 +432,14 @@ export async function makeUtxoEngine(
...
       if (id !== transaction.txid) {
-        throw new Error('broadcast response txid does not match original')
+        // The transaction is on the network at this point, so a mismatched
+        // response txid must not be reported as a send failure. Track the
+        // txid the network actually accepted, or the wallet would watch a
+        // transaction that never confirms.
+        log.warn(
+          `broadcast response txid mismatch: expected ${transaction.txid} received ${id}`
+        )
+        return { ...transaction, txid: id }
```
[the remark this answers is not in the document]
j0ntz: Fixed on branch head 3a1ab78: engineProcessor.broadcastTx now returns serverStates.broadcastTx()'s result (the txid the network answered with), so the mismatch branch is live and the wallet tracks the accepted txid. No unit harness reaches UtxoEngine.broadcastTx with a mocked server response today; noted as residual in the run report.

src/common/utxobased/engine/ServerStates.ts
```diff
@@ -393,12 +417,9 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
...
               })
               .catch((e?: Error) => {
+                broadcastErrors.push(e)
                 if (++bad === wsUris.length) {
-                  const msg = e != null ? `With error ${e.message}` : ''
-                  log.error(
-                    `broadcastTx fail: ${JSON.stringify(transaction)}\n${msg}`
-                  )
-                  reject(e)
+                  rejectClassified(e)
```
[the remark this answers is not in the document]
j0ntz: Half right, half fixed (3a1ab78): the Electrum stub refuses broadcastTx synchronously and provably never sent anything, so those failures are now excluded from the ambiguity determination (isNonRelayFailure; spec covers stub+rejection = rejected, all-stub = rejected, stub+timeout = ambiguous). A not-yet-connected blockbook entry stays ambiguous deliberately: its queued request can still send once the connection completes, so its timeout is genuine ambiguity, not a false positive.

src/common/utxobased/engine/broadcastError.ts
```diff
@@ -0,0 +1,58 @@
...
+  )
+
+export const classifyBroadcastFailure = (
+  errors: unknown[]
+): 'rejected' | 'ambiguous' => {
+  if (errors.length === 0) return 'ambiguous'
+  const relayCapable = errors.filter(error => !isNonRelayFailure(error))
+  return relayCapable.every(isExplicitBroadcastRejection)
+    ? 'rejected'
+    : 'ambiguous'
```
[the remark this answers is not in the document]
j0ntz: Valid, fixed on branch head 51e5b66: already-known rejections (transaction already in block chain, txn-already-in-mempool, txn-already-known) are now their own class (isAlreadyKnownRejection, gated on the Blockbook marker plus /alread/i). Any such error in the failure set RESOLVES the broadcast as a success with the submitted txid, since the server confirming it already has the transaction proves it reached the network; it never reaches the rejected/ambiguous classification. Spec covers the three message shapes plus the dust and transport negatives.

PR #458, opened 2026-09-03
Broadcast to every server at once, and stop leaking the NOWNodes key

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Engine half of the "send failed in the UI but the money moved" incident. Asana: https://app.asana.com/1/9976422036640/project/1213880789473005/task/1217135300337949

GUI companion (independent, neither blocks the other): EdgeApp/edge-react-gui#6190

Three commits, review fixups squashed in. **The first commit stands alone** and can be reviewed, or landed, without the rest.

---

#### 1. `Send the NOWNodes API key only to NOWNodes servers`

`ServerConfig` had a single type, `blockbook-nownode`, so when Edge's own Blockbook hosts were added in Dec 2024 (9962d7e) they were filed under it alongside the real NOWNodes entries. Every HTTP broadcast therefore sent the NOWNodes key to Edge's servers too.

They are not the same kind of server:

```
GET https://btc-wusa1.edge.app/api/v2   -> 200, no key needed
GET https://btcbook.nownodes.io/api/v2  -> 401 Unknown API_key
```

The type becomes `blockbook | blockbook-nownode`. The ten `edge.app` HTTP entries are now plain `blockbook`, the `nownodes.io` entries are unchanged, and the key is attached per target instead of to the whole batch.

This also fixes the no-key case. A missing `nowNodesApiKey` previously rejected the entire HTTP path with "Missing connection key for fallback servers", even though the public servers need no key. NOWNodes targets are now skipped with a warning and the public servers still carry the broadcast.

#### 2. `Reject socket requests when the socket fails`

Found in review. `promisifyWsMessage` wraps each request in a `Deferred` that was only ever resolved from inside its own generator. The socket reports every failure (error response, 30s expiry, socket close) by throwing into that generator, which had no try/catch, so the deferred stayed pending forever and the caller hung. That applied to `fetchInfo`, `fetchAddress`, `fetchAddressUtxos`, `fetchTransaction` and `ping`, not only `broadcastTx`. The generator now rejects its deferred on throw.

This has to precede the fan-out: once every socket attempt is counted, an all-fail broadcast with a cached socket could never settle without it.

#### 3. `Broadcast to every server at once`

The HTTP servers were used only when no WebSocket reported itself connected. A socket that looks connected but never answers `sendTransaction` therefore failed the whole broadcast, after the 30 second request timeout, without a single HTTP attempt being made. That is the shape of the incident behind this task.

The gate is worse than it looks: the NOWNodes WebSocket URI is docked 400 points at score load (`ServerScores.serverScoresLoad` penalises any URI carrying key params) and so rarely wins one of the two connection slots. HTTP is the realistic route to those servers, and it was reachable only when every socket was visibly down.

`broadcastTx` now fires in two waves. The first goes out immediately to every cached blockbook (sockets still connecting included, since a queued request transmits as soon as the socket opens) and to every `blockbook` HTTP server, which are Edge's own. The second wave, the `blockbook-nownode` HTTP servers, fires after `NOWNODES_BROADCAST_DELAY_MS` (2s), or as soon as every first-wave attempt has failed, whichever comes first. The first success resolves; it rejects only once every attempt has failed, logging each failure against its server. Every attempt is bounded by `BROADCAST_ATTEMPT_TIMEOUT_MS` (30s, mirroring the socket request expiry), so a server that accepts the connection and never answers cannot hold the broadcast open; the socket layer already expired its own requests, HTTP had no bound until Bugbot pointed it out.

The two waves are the answer to the review thread on unconditional HTTP. NOWNodes is a third party, so it only sees a transaction when Edge's own infrastructure has not already carried it. In the incident every socket was down and the HTTP fallback that saved the broadcast included Edge's own servers, so the first wave would very likely have succeeded on its own; the delay costs anything only in the double-failure case, and two seconds against the old thirty-second hang is nothing. `attempts` counts both waves up front, so an all-fail first wave cannot reject before the second has run.

### Testing

`test/common/utxobased/engine/ServerStates.spec.ts` (new) and `Blockbook.spec.ts`:

- broadcasting over every HTTP server when no sockets are cached
- first success wins when one server fails and another succeeds
- rejecting only after every attempt fails
- rejecting immediately when there is nothing to broadcast to
- NOWNodes is not contacted at all when Edge's own servers carry the broadcast (waits out the delay to prove the timer was cancelled)
- NOWNodes fires at once, not after the delay, when Edge's servers fail fast, with the `api-key` header on the NOWNodes request and **not** on the public one
- skipping NOWNodes but still using public servers when no key is configured
- rejecting when only NOWNodes servers exist and no key is configured
- `fetchInfo` rejects when the server answers with an error (commit 2)
- a real WebSocket that accepts `sendTransaction` and never answers: NOWNodes is held back for the delay, then the broadcast resolves through it well inside the 30s socket timeout
- a socket that refuses alongside failing Edge servers: NOWNodes fires at once and the broadcast resolves in under the delay
- an HTTP server that accepts and never answers times out and the broadcast rejects
- a refusing socket plus a hung HTTP server rejects instead of hanging, Bugbot's case
- the default attempt timeout equals the socket layer's 30 seconds
- a socket that refuses `sendTransaction` with no HTTP server rejects instead of hanging
- a socket that refuses alongside a failing HTTP server rejects instead of hanging

Full suite: 1261 passing at commit 1, 1262 at commit 2, 1270 at HEAD (re-run after the squash; tree identical to the reviewed head). `tsc` and `eslint` clean.

### Note for reviewers

This deliberately does not include the broadcast-failure classification from #455. Per the direction agreed for this task, the app now hard-fails an ambiguous broadcast and says so, rather than trying to infer whether the transaction landed. The `saveTx` "No addresses to process" fix from #455 is still valid and is not in this branch.

---

> [!NOTE]
> **High Risk**
> Changes core transaction broadcast paths, third-party exposure timing, and API key handling—directly affects whether sends succeed or fail and whether txs leak to NOWNodes early.
> 
> **Overview**
> Fixes UTXO sends that could **fail in the UI while the transaction still propagated**, by changing how `broadcastTx` reaches Blockbook.
> 
> **`broadcastTx` now fans out in two waves.** The first wave immediately hits every cached Blockbook WebSocket (including sockets still connecting) and every Edge `blockbook` HTTP host. NOWNodes `blockbook-nownode` HTTP targets run in a second wave after **2 seconds**, or **immediately** once the first wave has all failed—whichever comes first. The **first success wins**; the call rejects only when **every** attempt fails or hits a **30s** per-attempt timeout (HTTP previously could hang indefinitely). This replaces the old rule that skipped HTTP whenever any socket looked connected, which let one silent socket block fallback entirely.
> 
> **Server typing and credentials:** `ServerConfig` is split into `blockbook` (no headers) vs `blockbook-nownode` (`api-key` only on NOWNodes). Edge `*.edge.app` HTTP entries across coin configs are reclassified as `blockbook`, so the NOWNodes key is no longer sent to Edge. Without a key, NOWNodes servers are skipped with a warning while public Edge HTTP still broadcasts.
> 
> **Socket hang fix:** `Blockbook.promisifyWsMessage` now **rejects** when the socket throws (errors, timeout, close), so broadcast and other WS calls settle instead of pending forever—required for counting all broadcast attempts.
> 
> New **`ServerStates.broadcastTx`** tests cover wave timing, header placement, timeouts, and socket+HTTP combinations; **`Blockbook.spec`** adds error-response rejection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b1d77cf4647a8f29f3d1db7b47d56c0675529b81. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/common/utxobased/engine/ServerStates.ts
+++ b/src/common/utxobased/engine/ServerStates.ts
@@ -39,6 +39,11 @@ interface ServerStateConfig {
   pluginInfo: PluginInfo
   pluginState: PluginState
   walletInfo: SafeWalletInfo
+  /**
+   * Upper bound on any single broadcast attempt. Defaults to the same 30
+   * seconds the socket layer uses for a request. Exposed for tests.
+   */
+  broadcastTimeoutMs?: number
 }
 
 export interface ServerStates {
@@ -92,6 +97,42 @@ interface ServerStatesCache {
   [uri: string]: ServerState
 }
 
+/**
+ * How long broadcastTx gives the first wave (our sockets and Edge's own HTTP
+ * servers) before it also sends the transaction to the NOWNodes HTTP servers.
+ * NOWNodes is a third party, so it should only see a transaction when our own
+ * infrastructure has not already carried it. A first wave that fails outright
+ * does not wait this long; the second wave fires as soon as it has all failed.
+ */
+export const NOWNODES_BROADCAST_DELAY_MS = 2000
+
+/**
+ * Every broadcast attempt is counted toward "all attempts failed", so every
+ * attempt must settle. Sockets already expire a request after 30 seconds
+ * (Socket.ts); HTTP has no such bound of its own, and a server that accepts
+ * the connection and never answers would otherwise hold the whole broadcast
+ * open. This mirrors the socket figure.
+ */
+export const BROADCAST_ATTEMPT_TIMEOUT_MS = 30000
+
+const withTimeout = async <T>(
+  promise: Promise<T>,
+  ms: number,
+  message: string
+): Promise<T> => {
+  let timer: ReturnType<typeof setTimeout> | undefined
+  const timeout = new Promise<never>((resolve, reject) => {
+    timer = setTimeout(() => {
+      reject(new Error(message))
+    }, ms)
+  })
+  try {
+    return await Promise.race([promise, timeout])
+  } finally {
+    if (timer != null) clearTimeout(timer)
+  }
+}
+
 export function makeServerStates(config: ServerStateConfig): ServerStates {
   const {
     engineEmitter,
@@ -103,6 +144,7 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
     walletInfo
   } = config
   const { serverConfigs = [] } = pluginInfo.engineInfo
+  const { broadcastTimeoutMs = BROADCAST_ATTEMPT_TIMEOUT_MS } = config
   log('Making server states')
 
   const serverStatesCache: ServerStatesCache = {}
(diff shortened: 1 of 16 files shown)


j0ntz (commented):
Reviewed alongside EdgeApp/edge-react-gui#6190; the findings are inline.

Nit: the commit subject `Broadcast to every server at once instead of gating HTTP on the sockets` is 71 characters, over Edge's 50-character limit. `Broadcast to every server at once` is 33 and the body already carries the rest. (`Send the NOWNodes API key only to NOWNodes servers` is exactly 50 and fine.)

src/common/utxobased/engine/ServerStates.ts
```diff
@@ -366,112 +366,127 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
...
-          }
+        //
+        // WebSocket blockbook servers (every cached connection):
+        //
+        for (const uri of Object.keys(serverStatesCache)) {
+          const { blockbook } = serverStatesCache[uri]
+          if (blockbook == null) continue
+          attempts++
+          blockbook
+            .broadcastTx(transaction)
```
j0ntz: **A failing WebSocket attempt never settles, so this promise can hang instead of rejecting.**

`blockbook.broadcastTx` goes through `promisifyWsMessage` (Blockbook.ts), which awaits a `Deferred` that is only ever resolved from inside its own generator. Every failure path in `Socket.ts` delivers the error as `request.generator.throw(err).catch(e => log.error(e.message))`: the error-response branch in `onMessage`, the 30s expiry in `onTimer`, and `onSocketClose`. That generator has no try/catch, so the rejection is logged and the deferred is left pending forever. No `onFailure` ever runs for that attempt, `failures` never reaches `attempts`, and the promise settles neither way.

```mermaid
sequenceDiagram
    participant SS as ServerStates.broadcastTx
    participant Sock as Socket blockbook
    participant Http as HTTP servers
    SS->>Sock: broadcastTx, attempts = 1
    SS->>Http: sendtx, attempts = 2
    Http-->>SS: rejects, onFailure, failures = 1
    Sock->>Sock: 30s expiry, generator.throw
    Sock->>Sock: Deferred left pending, no onFailure
    SS->>SS: failures 1 < attempts 2, never settles
```

The deferred bug is pre-existing, but the denominator change is what exposes it. Before this commit, the fully-disconnected case took the `!isAnyBlockbookConnected` branch, skipped sockets entirely and rejected off the HTTP count, so a genuinely failed broadcast surfaced as a failure. Sockets are now always counted, so an all-fail broadcast with any entry in `serverStatesCache` never rejects. Paired with EdgeApp/edge-react-gui#6190 that is a locked slider spinning forever with no message at all, since the scene's `finally` never runs.

The new spec covers socket-hangs-but-HTTP-succeeds; it is the all-fail variant that hangs. Either bound each attempt before counting it, or have `promisifyWsMessage` reject its deferred when the generator throws.

src/common/utxobased/engine/ServerStates.ts
```diff
@@ -366,112 +366,127 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
...
-            log.warn('Falling back to NOWNode server broadcast over HTTP:', uri)
-
-            // HTTP Fallback
-            io.fetchCors(`${uri}/api/v2/sendtx/`, {
-              method: 'POST',
-              headers: {
-                'api-key': nowNodesApiKey
-              },
-              body: transaction.signedTx
+        for (const { uri, headers } of httpTargets) {
```
j0ntz: Every send now discloses the transaction to the NOWNodes HTTP endpoint even when the sockets and Edge's own servers are healthy, where HTTP was previously a fallback only. For BTC that is `btc-wusa1.edge.app`, `btc-eu1.edge.app` and `btcbook.nownodes.io` on every broadcast, unconditionally.

Unblocking HTTP is the right call; making it unconditional on the happy path is a separate product decision that the PR description does not call out. Worth confirming it is intended, or staggering the third-party targets behind a short delay so they only fire when the first wave has not resolved.

j0ntz (changes requested):
Correcting the verdict on my earlier review: those findings are blocking, not advisory.

The broadcast hang is the one that gates this. An all-fail broadcast with any entry in `serverStatesCache` settles neither way, and paired with EdgeApp/edge-react-gui#6190 the send scene has no error to show and no `finally` to unlock the slider.

src/common/utxobased/engine/ServerStates.ts
peachbits: Confirmed, and it is wider than broadcast: the generator in promisifyWsMessage has no try/catch around its yield, so every generator.throw from Socket.ts (error response, 30s expiry, socket close) left the deferred pending for fetchInfo, fetchAddress, fetchAddressUtxos, fetchTransaction and ping as well.

Fixed at the root in 40a585d ("Reject socket requests when the socket fails"), now the second commit so the fan-out lands on top of it: the generator catches the throw and rejects its deferred, with a Blockbook spec asserting fetchInfo rejects on an error response. Fixup 0d32978 adds the two all-fail specs you asked for, a socket that refuses sendTransaction with no HTTP server, and one that refuses alongside a failing HTTP server. Both settle by rejection well inside the 30s timeout.

peachbits:
Subject reworded to `Broadcast to every server at once` (33 chars). The branch is now four commits: the key fix first as requested, then the promisifyWsMessage rejection fix found via your hang finding, then the fan-out, then a fixup with the all-fail specs. The unconditional-HTTP question on the other thread is with Matthew as a product decision; I will follow up there.

src/common/utxobased/engine/ServerStates.ts
peachbits: Agreed, and decided with Matthew: the third-party targets are now staggered. Fixup 32c5992 splits the fan-out into two waves. Sockets and the `blockbook` (Edge) HTTP servers fire immediately; the `blockbook-nownode` servers fire after NOWNODES_BROADCAST_DELAY_MS (2s), or at once when every first-wave attempt has already failed, so a fast failure does not sit out the delay. A first-wave success cancels the timer, so on the happy path NOWNodes never sees the transaction. The `blockbook | blockbook-nownode` split from the first commit is what makes the waves fall out cleanly.

Specs cover all three shapes: not contacted when Edge's servers carry it (waits out the timer to prove cancellation), held back for the delay behind a silent socket, and fired at once when the first wave fails fast. The fan-out commit message and PR description now describe the two-wave behavior.

src/common/utxobased/engine/ServerStates.ts
```diff
@@ -366,112 +375,168 @@ export function makeServerStates(config: ServerStateConfig): ServerStates {
...
-          // settings:
-          // const httpUris = pluginState.getLocalServers(Infinity, [
-          //   /^http(?:s)?:/i
-          // ])
-
-          const { nowNodesApiKey } = initOptions
-          const nowNodeUris = serverConfigs
-            .filter(config => config.type === 'blockbook-nownode')
-            .map(config => config.uris)
-            .flat(1)
```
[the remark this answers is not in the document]
peachbits: Legitimate, and fixed in ce3c96a. Every attempt is now raced against BROADCAST_ATTEMPT_TIMEOUT_MS (30s, mirroring the socket layer's request expiry), overridable via ServerStateConfig.broadcastTimeoutMs for tests, so a server that accepts the connection and never answers settles as a failure instead of pinning `failures` below `attempts`. The underlying fetch is left to finish on its own; only the attempt is settled. Specs added for a hung HTTP server on its own, for your exact case of a refusing socket plus a hung HTTP server (rejects in well under the socket timeout), and for the default matching the socket figure. 1270 passing.

j0ntz (approved):

